PR #12, opened 2014-12-29
Validation reengineering

0xDiBa:


src/certAnalyser/analyserPool.go
```diff
@@ -186,69 +198,135 @@ func worker(msgs <-chan amqp.Delivery, es *elastigo.Conn) {
...
+
+		return false
+	}
+
+}
+
+//Returns the first parent found for a certificate in a given certificate list ( does not verify signature)
+func getFirstParent(cert *x509.Certificate, certs []*x509.Certificate) *x509.Certificate {
+	for _, c := range certs {
+		if cert.Issuer.CommonName == c.Subject.CommonName { //TODO : consider changing this check with validating check
```
jvehent: Does this need to verify signatures, considering the chain is already validated by Go?

src/certAnalyser/analyserPool.go
```diff
@@ -584,12 +681,40 @@ func main() {
 
 	failOnError(err, "Failed to register a consumer")
 
+	//Register truststores
+	for i, name := range conf.TrustStores.Name {
+
+		poolData, e := ioutil.ReadFile(conf.TrustStores.Path[i])
```
jvehent: replace `e` with `err` for consistency

jvehent:
Looks good. I'll merge and we can fix the minor comments in master.

src/certAnalyser/analyserPool.go
0xDiBa: Not really.
I was just laying out some thoughts. 
Have to get rid of non-useful comments :)

PR #49, opened 2015-10-21
Update to the DB schema to support chains of trust

jvehent:
This schema should support multiple chain of trusts per certificate, as well as updating the status of a chain of trust over time.
schema

r? @0xDiBa @marumari

### DIFF
--- a/modules/postgresmodule/schema.sql
+++ b/modules/postgresmodule/schema.sql
@@ -51,6 +51,7 @@ CREATE TABLE scans  (
 	replay 				        integer NULL,
 	has_tls						bool NOT NULL,
 	cert_id		              	integer references certificates(id) NOT NULL,
+    trust_id                    integer references trust(id) NOT NULL,
 	is_valid                   	bool NULL,
 	completion_perc				integer NULL,
 	validation_error           	varchar NULL,
--- a/modules/postgresmodule/schema.sql
+++ b/modules/postgresmodule/schema.sql
@@ -1,27 +1,3 @@
-CREATE TABLE scans  (
-	id                         	serial primary key,
-	time_stamp	           		timestamp NOT NULL,
-	target						varchar NOT NULL,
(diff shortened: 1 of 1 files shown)


marumari:
I'm a bit skeptical on all the trusted_ fields; are those going to be regenerated constantly?

Also, shouldn't the scan be saying which trust chain that it used at the time of the scan?

jvehent:
> I'm a bit skeptical on all the trusted_ fields; are those going to be regenerated constantly?

The scanner will rerun the tests and only store a new row in the `trust` table if the results are different from the current row. So, yes, the data is regenerated constantly but no it isn't stored in database as long as it doesn't change.

> Also, shouldn't the scan be saying which trust chain that it used at the time of the scan?

That's a good point. I'll add a field.

jvehent:
Schema updated with the reference to the chain of trust used.
schema

marumari:
That looks a lot better to me.  I'm just questioning if we need to even have the trusted_ fields at all?  Surely when we view a single certificate, we can just run the trusted tests then?  I can only really see it useful to store if we're running big queries constantly against those fields.

I guess honestly it's not a big deal either way, since it's just a boolean.  :)

In any case, r+.

jvehent:
The motivation behind storing the truststore validation is to analyze the impact of a truststore update on certificates, before and after the change. Without storing those fields, we would need to keep versions of each truststore and rerun the validation against all of them. I think it's worth storing that data, giving it's just a few booleans. 

Not that we may still need to store versions of truststore in DB anyway. That's probably the next thing to add to the schema, but it's not a blocker to move forward on the current issue.

0xDiBa:
I think that eventually, and for "research" purposes, we are going to derive current truststore versions from the in_xxx_root_store variables of CA certificates. Coming to the point of storing versions of each truststore will help us get a historical overview of certificates in truststores.

I think we are good to go ( and merge ) for now.

PR #52, opened 2015-11-16
Reorganize repository structure

jvehent:
I changed a bunch of directories around, to make the repository more go-ish. I also changed some of the goroutine logic in the tlsobs-scanner (formely tlsObserver) to be more linear.
Both tlsobs-scanner and tlsobs-api build with go get, but I have not tested deployment or functionality.

@0xDiBa: r?


0xDiBa:
Looks good. :+1: 
I like your naming better than mine.
tlsobs-scanner is going to change more once the additional workers come in...

jvehent:
When you mean `change`, do you mean the code or the naming?
The code has to change for sure, since I removed the rabbitmq support and just put a dumb channel in there that doesn't do anything. We'll have to replace that with the SELECT FOR UPDATE logic that feeds new scan IDs into the incomingScans channel.

0xDiBa:
The code of course ;)

0xDiBa:
Since it has not been merged yet, I made a patch here.
Messaging through postgres now works.

jvehent:
awesome! thanks!

PR #55, opened 2015-11-23
Remove ts from conf

0xDiBa:
Explicit truststore configuration.
Only path allowed for certain truststores.

### DIFF
--- a/certificate/analyser.go
+++ b/certificate/analyser.go
@@ -32,15 +32,12 @@ func Setup(c config.Config, DB *pg.DB) {
 
 		switch name {
 		case ubuntu_TS_name:
-
 			path = ts.UbuntuTS
 
 		case mozilla_TS_name:
-
 			path = ts.MozillaTS
 
 		case microsoft_TS_name:
-
 			path = ts.MicrosoftTS
 
 		case apple_TS_name:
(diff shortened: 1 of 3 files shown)


certificate/analyser.go
```diff
@@ -19,17 +19,51 @@ import (
...
+
+			path = ts.MicrosoftTS
+
+		case apple_TS_name:
+			path = ts.AppleTS
+
+		case android_TS_name:
+			path = ts.AndroidTS
+
+		default:
```
jvehent: stylistic: please remove the newlines, it's just as readable without them ;)

certificate/analyser.go
```diff
@@ -19,17 +19,51 @@ import (
...
+
+			log.WithFields(logrus.Fields{
+				"tsname": name,
+			}).Warning("Wrong slice name entry. This should not be there")
```
jvehent: need a more explicit error message, such as
`Truststore %q is not a valid name", name`

jvehent:
I like it. there's just a couple nits, and it's good to merge!

0xDiBa:
Need to work on my styling skills ;)
Merging...

0xDiBa:
Did not realise that.
Add work to my common understanding skill.

Coming right up on my next commit ;)

PR #59, opened 2015-11-25
add yaml description of elasticbeanstalk application

jvehent:
@mostlygeek : when you get a chance, could you take a look at the attached yaml file and let me know if it makes sense to you? It's an adapted copy/paste from the ideatown template.


mostlygeek:
The YAML file looks ok to me. Though it's sort of orphaned without all the other tools in our ops repo. :)

jvehent:
Thanks! So given this file, could you guide me through the next steps?(might be easier do to that over Vidyo next week if you prefer)mozilla/tls-observatory wrote:The YAML file looks ok to me. Though it's sort of orphaned without all the other tools in our ops repo. :)

—Reply to this email directly or view it on GitHub.


  
  


Sent Using Firefox OS

PR #61, opened 2015-11-28
Api upgrade

0xDiBa:
Just for review, so far.
Main functionality is there but needs some refinements and cljs incorporation.

### DIFF
--- a/certificate/certificate.go
+++ b/certificate/certificate.go
@@ -23,48 +23,48 @@ const android_TS_name = "Android"
 type Certificate struct {
 	ScanTarget             string                    `json:"scanTarget,omitempty"`
 	IPs                    []string                  `json:"ips,omitempty"`
-	Version                float64                   `json:"version"`
-	SignatureAlgorithm     string                    `json:"signatureAlgorithm"`
-	Issuer                 Issuer                    `json:"issuer"`
-	Validity               Validity                  `json:"validity"`
-	Subject                Subject                   `json:"subject"`
-	SubjectPublicKeyInfo   SubjectPublicKeyInfo      `json:"subjectPublicKeyInfo"`
-	X509v3Extensions       Extensions                `json:"x509v3Extensions"`
-	X509v3BasicConstraints string                    `json:"x509v3BasicConstraints"`
-	CA                     bool                      `json:"ca"`
-	Analysis               interface{}               `json:"analysis"` //for future use...
-	ParentSignature        []string                  `json:"parentSignature"`
-	ValidationInfo         map[string]ValidationInfo `json:"validationInfo"`
-	FirstSeenTimestamp     string                    `json:"firstSeenTimestamp"`
-	LastSeenTimestamp      string                    `json:"lastSeenTimestamp"`
-	Hashes                 Hashes                    `json:"hashes"`
-	Raw                    string                    `json:"Raw"`
+	Version                float64                   `json:"version,omitempty"`
+	SignatureAlgorithm     string                    `json:"signatureAlgorithm,omitempty"`
+	Issuer                 Issuer                    `json:"issuer,omitempty"`
+	Validity               Validity                  `json:"validity,omitempty"`
+	Subject                Subject                   `json:"subject,omitempty"`
+	SubjectPublicKeyInfo   SubjectPublicKeyInfo      `json:"subjectPublicKeyInfo,omitempty"`
+	X509v3Extensions       Extensions                `json:"x509v3Extensions,omitempty"`
+	X509v3BasicConstraints string                    `json:"x509v3BasicConstraints,omitempty"`
+	CA                     bool                      `json:"ca,omitempty"`
+	Analysis               interface{}               `json:"analysis,omitempty"` //for future use...
+	ParentSignature        []string                  `json:"parentSignature,omitempty"`
+	ValidationInfo         map[string]ValidationInfo `json:"validationInfo,omitempty"`
+	FirstSeenTimestamp     time.Time                 `json:"firstSeenTimestamp"`
+	LastSeenTimestamp      time.Time                 `json:"lastSeenTimestamp"`
+	Hashes                 Hashes                    `json:"hashes,omitempty"`
+	Raw                    string                    `json:"Raw,omitempty"`
 	Anomalies              string                    `json:"anomalies,omitempty"`
 }
 
 type Issuer struct {
-	Country      []string `json:"c"`
-	Organisation []string `json:"o"`
-	OrgUnit      []string `json:"ou"`
-	CommonName   string   `json:"cn"`
+	Country      []string `json:"c,omitempty"`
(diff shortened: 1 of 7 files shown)


jvehent:
will review tomorrow

certificate/certificate.go
```diff
@@ -116,8 +116,21 @@ type TrustStore struct {
...
+	ID               int64
+	CertID           int64
+	IssuerID         int64
+	Timestamp        time.Time
+	TrustUbuntu      bool
+	TrustMozilla     bool
+	TrustedMicrosoft bool
+	TrustedApple     bool
+	TrustedAndroid   bool
+	Current          bool
```
jvehent: what's `current` ?

certificate/certificate.go
```diff
@@ -223,6 +236,27 @@ func (c Certificate) GetBooleanValidity() (trusted_ubuntu, trusted_mozilla, trus
 	return
 }
 
+// GetValidityMap converts boolean validity variables to a validity map.
+func GetValidityMap(trusted_ubuntu, trusted_mozilla, trusted_microsoft, trusted_apple, trusted_android bool) map[string]ValidationInfo {
```
jvehent: Can you just pass a `Trust` variable instance of all those?

certificate/certificate.go
```diff
@@ -223,6 +236,27 @@ func (c Certificate) GetBooleanValidity() (trusted_ubuntu, trusted_mozilla, trus
...
+	vAndroid := ValidationInfo{IsValid: trusted_android}
+
+	m := make(map[string]ValidationInfo)
+
+	m[ubuntu_TS_name] = vUbuntu
+	m[mozilla_TS_name] = vMozilla
+	m[microsoft_TS_name] = vMicrosoft
+	m[apple_TS_name] = vApple
+	m[android_TS_name] = vAndroid
+
```
jvehent: I'm a bit confused... is this just converting the input trusted booleans into a map without doing any other computation? If so, that looks like something that should be handled with structs instead.

certificate/database.go
```diff
@@ -14,6 +14,12 @@ import (
 
 var db *pg.DB
 
+// SetDB sets the pointer to the database.
+// Probably some bad design decisions led to this... :)
+func SetDB(DB *pg.DB) {
+	db = DB
+}
```
jvehent: This is smelling wrong... very wrong... I would agree with your comment here.

certificate/database.go
```diff
@@ -314,7 +320,7 @@ func GetCertwithSHA1Fingerprint(sha1 string) (*Certificate, error) {
 
 // GetCertwithID fetches a certain certificate from the database.
 // It returns a pointer to a Certificate struct and any errors that occur.
-func GetCertwithID(id string) (*Certificate, error) {
+func GetCertwithID(certID int64) (*Certificate, error) {
```
jvehent: `GetCertByID` would be a better name... just saying :wink:

certificate/database.go
```diff
@@ -459,6 +473,46 @@ func getCurrentTrust(certID, issuerID int64) (int64, error) {
...
+
+		if err == sql.ErrNoRows {
+			return -1, nil
+		} else {
+			return -1, err
+		}
+	}
+
+	return trustID, nil
+}
```
jvehent: All those unnecessary newlines.... I can taste those delicious Belgium beers already!

certificate/database.go
```diff
@@ -459,6 +473,46 @@ func getCurrentTrust(certID, issuerID int64) (int64, error) {
...
+	var trustID int64
+
+	row := db.QueryRow("SELECT id FROM trust WHERE cert_id=$1 AND is_current=TRUE", certID)
+
+	err := row.Scan(&trustID)
+
+	if err != nil {
+
+		if err == sql.ErrNoRows {
+			return -1, nil
```
jvehent: is `-1` a convention for error code? might be better to define those as constants instead.

connection/connection.go
```diff
@@ -33,23 +33,23 @@ type CipherscanCiphersuite struct {
...
-	FirstSeenTimestamp string                       `json:"firstSeenTimestamp"`
-	LastSeenTimestamp  string                       `json:"lastSeenTimestamp"`
-	ServerSide         bool                         `json:"serverside"`
-	CipherSuites       map[string]StoredCiphersuite `json:"cipherscanCiphersuite"`
-	CurvesFallback     bool                         `json:"curvesFallback"`
+	ScanIP             string                       `json:"scanIP,omitempty"`
+	FirstSeenTimestamp string                       `json:"firstSeenTimestamp,omitempty"`
+	LastSeenTimestamp  string                       `json:"lastSeenTimestamp,omitempty"`
+	ServerSide         bool                         `json:"serverside,omitempty"`
+	CipherSuites       map[string]StoredCiphersuite `json:"cipherscanCiphersuite,omitempty"`
```
jvehent: `CipherSuites` should be singular: `CipherSuite`
The json name shouldn't mention `cipherscan` at all, unless you're returning not just the ciphersuite, but the entire cipherscan json, in which case I would call this `cipherscanOutput` or something similar

tlsobs-api/handlers.go
```diff
@@ -69,17 +80,74 @@ func ResultHandler(w http.ResponseWriter, r *http.Request) {
...
+	val, ok := context.GetOk(r, dbKey)
+	if !ok {
+		log.Error("Could not find db in request context")
+		err = errors.New("Could not access database.")
+		return
+	}
+
+	db := val.(*pg.DB)
+
+	scanIDStr := r.FormValue("scan_id")
```
jvehent: same comment as for the certificate endpoint: we're already in the `scan` api namespace, so just calling it `id` is sufficient, since we already now we're requesting a scan

tlsobs-api/handlers.go
```diff
@@ -95,25 +163,67 @@ func CertificateHandler(w http.ResponseWriter, r *http.Request) {
...
+		log.WithFields(logrus.Fields{
+			"scan_id": certID,
+			"error":   err.Error(),
+		}).Error("Could not get cert from database")
+
+		err = errors.New("Something went wrong :\\")
+		return
+	}
+
+	jsScan, err := json.MarshalIndent(cert, "", "	")
```
jvehent: same as before: don't return indented json, let the client take care of that.

jvehent:
I might have a comment or two :stuck_out_tongue_closed_eyes:

certificate/database.go
0xDiBa: I wanted to handle the NoRows error inside the database functions and -1 without an error seemed like good way to say not found without having the caller handling an error.
It might be better to have some constants though.

certificate/database.go
0xDiBa: I will have to review all the code :smile: , or I might as well make an offer to buy you the Duvel brewery :beers:

PR #62, opened 2015-11-29
General restructure

0xDiBa:
@jvehent Do you think this package structuring is somewhat cleaner?


jvehent:
It this based of pr #61 ? I can see previous comments in there...
Are you talking specifically about d22d7f6 that uses method from the database package? If so, that looks clean to me.

0xDiBa:
Yeah it is based on that PR.
I didn't realise that github added all the previous comments. Sorry about that.

I was indeed talking specifically about that last commit. 
I think moving all these methods to the database package 
and moving the certificate scanning methods to the scanner module looks cleaner than the previous 
"thing" :smile:

jvehent:
That seems like a good idea. Make sure to merge #61 first though.mozilla/tls-observatory wrote:Yeah it is based on that PR.
I didn't realise that github added all the previous comments. Sorry about that.

I was indeed talking specifically about that last commit. 
I think moving all these methods to the database package 
and moving the certificate scanning methods to the scanner module looks cleaner than the previous 
"thing"  

—Reply to this email directly or view it on GitHub.


  
  


Sent Using Firefox OS

PR #65, opened 2015-12-04
Set completion to 100 when a target doesn't support TLS

jvehent:
@0xDiBa : r?


0xDiBa:
Seems reasonable :sweat_smile:

PR #70, opened 2015-12-06
Several database improvements

jvehent:
The support for SSL in postgres is there, but not enabled by default because I'm getting a cert mismatch from RDS.

I also added a small Go script to fix the validity of existing certs in the database, by reparsing the raw certs.

@0xDiBa: r?


0xDiBa:
Looks good.

PR #76, opened 2015-12-14
Pgmessagingv2

### LINKED ISSUE #64: Catch up on missed scans
If the scanners crash, they need to pick up incomplete work. One way to do that is to periodically run this query:
```sql
select pg_notify('scan_listener', ''||id ) from scans where completion_perc=0;
```
which will resend scan notifications to target that are still at 0% completion.

0xDiBa:
Should I merge and close  #64?

### DIFF
--- a/database/messaging.go
+++ b/database/messaging.go
@@ -69,26 +69,22 @@ func (db *DB) RegisterScanListener(dbname, user, password, hostport, sslmode str
 	}()
 
 	go func() {
-		fiveminticker := time.NewTicker(1 * time.Minute)
-		unackedQuery := fmt.Sprintf("select pg_notify('%s', ''||id ) from scans where ack=FALSE and timestamp < NOW() - INTERVAL '30 seconds'", listenerName)
-		zerocomplQuery := "update scans set ack=false where completion_perc=0 and timestamp < NOW() - INTERVAL '1 minute'"
 		for {
-			select {
-			case <-fiveminticker.C:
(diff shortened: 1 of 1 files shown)


database/messaging.go
```diff
@@ -66,6 +68,26 @@ func (db *DB) RegisterScanListener(dbname, user, password, hostport, sslmode str
...
+	go func() {
+		for {
+			_, err := db.Exec("update scans set ack=false where completion_perc=0 and timestamp < NOW() - INTERVAL '1 minute'")
+			if err != nil {
+				log.WithFields(logrus.Fields{
+					"error": err,
+				}).Error("Could not run zero completion update query")
+			}
+
+			_, err = db.Exec(fmt.Sprintf("select pg_notify('%s', ''||id ) from scans where ack=FALSE and timestamp < NOW() - INTERVAL '30 seconds'", listenerName))
```
jvehent: I think it would be wise to add a `LIMIT` to this query, because the first time we run it, it will notify ~4 million scans!!!
What about `LIMIT 1000`, which should be enough for 5 minutes of activity?

0xDiBa: Wise words...
Although I would be curious to see what would happen in that case :wink:

jvehent: lots and lots of very busy processes, autoscaling reaching a max of 6 instances, database load going 100% until all systems reach a halt and I need to SSH in to stop all of it.

been there, done that :sob:

PR #77, opened 2015-12-15
issuer and subject are now jsonb fields

0xDiBa:
The proposed change breaks the current database schema as the issuer and subject fields are going from varchar to jsonb.

I think dropping and recreating these columns is the best thing we can do. 
Should I create a script to go through the whole certificates table, parse certificates and save the new form of the issuer and subject fields?

### DIFF
--- a/certificate/certificate.go
+++ b/certificate/certificate.go
@@ -486,9 +486,15 @@ func (i Issuer) String() (str string) {
 		str += "C=" + strings.Join(i.Country, ", C=")
 	}
 	if len(i.Organisation) > 0 {
+		if str != "" {
+			str += ", "
+		}
 		str += "O=" + strings.Join(i.Organisation, ", O=")
 	}
 	if len(i.OrgUnit) > 0 {
+		if str != "" {
+			str += ", "
+		}
 		str += "OU=" + strings.Join(i.OrgUnit, ", OU=")
 	}
 	if str != "" {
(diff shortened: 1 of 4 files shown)


jvehent:
We can drop and recreate, but you should write a script that reparse existing certificates using the raw_cert to repopulate the columns.

If possible, we should only do one pass over the certificates table and add the `domains` column as well. Can you add a commit for that?

0xDiBa:
Yeah i had that in mind too.
They are in the same issue.
The domains column should include the sucject common name and the subject
alternative names right?

jvehent:
> The domains column should include the sucject common name and the subject alternative names right?
Yes, as a string, comma separated

0xDiBa:
Below commands are probably going to be needed for the database: 
```ALTER TABLE certificates DROP COLUMN subject;
ALTER TABLE certificates DROP COLUMN issuer;
ALTER TABLE certificates ADD COLUMN subject jsonb;                                                                                                                                                
ALTER TABLE certificates ADD COLUMN issuer jsonb;
ALTER TABLE certificates ADD COLUMN domains varchar;```

I also changed the offset logic from the fixvalidity.go script a little bit because I could not parse
all the certificates in on pass.

jvehent:
It's working. I think the API needs to be updated to serve the new `domains` column?

0xDiBa:
I intentionally did not include it.
It's definitely info that is useful inside the database but on the other
side
Do we want to transfer duplicate info again over the net?
If it is needed it can easily be added to the certificate struct and be
served through the api.

jvehent:
That's reasonable, we can omit it from the API and let people look into subject and subjectAlternateName.

So, this looks good, I'm merging it :+1:

jvehent:
quick update: it tooks ~30 hours, but the production DB is now entirely converted to the new format! :beers:

0xDiBa:
That took a lot...
We should have done the parsing and storage in separate goroutines for each cert.
I did not know we had that many certs indexed already ;)

PR #85, opened 2015-12-22
Evaluation worker & scanners limits rework

jvehent:
do not merge yet, testing in staging first

This set of patches improves the evaluation worker and add some stricter controls on the number of scanners that run in parallel. We also store a number column in the `scans` table to track a count of retries on scans that don't succeeds.

This last change requires a DB change:
```sql
ALTER TABLE scans ADD COLUMN attempts integer;
```

### DIFF
--- a/conf/api.cfg
+++ b/conf/api.cfg
@@ -3,7 +3,7 @@
 # export TLSOBS_API_ENABLE=1
 Enable        = on
 
-GoRoutines    = 10
+MaxProc       = 10
 # IP:PORT of the postgres database, can be superseded by the environment variable
 # export TLSOBS_POSTGRES="127.0.0.1:5432"
 Postgres      = "127.0.0.1:5432"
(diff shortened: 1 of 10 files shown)


jvehent:
I'm not happy with the evaluation worker part of this. It's still way too strict, and even too verbose now that each missing/extra cipher for each evaluation level is reported.
This needs more work before getting merged.

On the other hand, the `attempts` logic seems to work as expected. Example here: http://apistage201512031437-2sjrnmheum.elasticbeanstalk.com//api/v1/results?id=1038

0xDiBa:
The limit on attempts seems nice.
I think that the "requirements" of the evaluation worker have grown. 
Do you think a redesign of the logic could solve any of the current problems?

jvehent:
Yeah, I think we'll need to rethink our approach sooner rather than later. But that can wait a little while. I want to work on recurring scans & alerting first.

jvehent:
Deployed this in production. It's working well. Had to fix the attempts manually because it would break the api otherwise.
```sql
update scans set attempts=1 where attempts is null;
```

PR #88, opened 2015-12-29
Tlsobs runner

### LINKED ISSUE #83: Create a scan-runner for recurring scans * alerting
We want to scan specific targets regularly, and send notification to various places. A new API client called `scan-runner` should do this.

Configuration should take a list of:
* target site
* evaluation
* notification

For example:
* site mozilla.org should evaluate to mozilla configuration `old` and notify someaddress@mozilla.com if evaluation fails
* site accounts.firefox.com should evaluate to `expiration > 30 days` and notify someotheraddress@mozilla.com if evaluation fails

jvehent:
This PR fixes #83 by implementing an API client that schedules scans and evaluates the results against specific assertions. This approach avoid hardcoding assertions for target sites in analysis worker, while still using those to do the heavy lifting.

IRC support is not yet implemented, but I think this is good enough to go for now.

@0xDiBa & @marumari : feedback?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,10 @@ env:
     - DOCKER_USERNAME=jvehent
     - secure: "LwcQZ/M+4M+HJf6b/bKcGdmuTcT4XeN5MVuBSy34ag79/Zz3UinwBpW3peOhFNizJUbMVzPJJ7FGMWeUouq6lCcjjQZAIBnPWx/bhP0z7vIO8t2+XqKV47D3HqPOokqEdeKHvzr4xvfKGMBy+VMCWeuFI7Z3BLCqtWSQYDjAMTg="
 script:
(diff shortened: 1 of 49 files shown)


0xDiBa:
Looks good :+1: 
Would it make sense, in a next step, to add all this logic to the backend 
and be able to have "assertion groups" ( for example AssertNotAfter-NotBefore ) in which replayed scans can be added and analysed automatically without having to update configuration and running a separate client?

jvehent:
> Would it make sense, in a next step, to add all this logic to the backend and be able to have "assertion groups" ( for example AssertNotAfter-NotBefore ) in which replayed scans can be added and analysed automatically without having to update configuration and running a separate client?

I think the issue is not only asserting criteria, but also doing something with the result of the assertion. I don't know if running recurring scans and sending email alerts in the core of the observatory is a good idea. It may be preferable to let people run their own `tlsobs-runner` with their own configuration on their own machine. I don't know what the right answer is, it's just a thought.

PR #102, opened 2016-02-15
TLS configuration from static resource

0xdiba:
The evaluation worker gets the configuration from https://statics.tls.security.mozilla.org/server-side-tls-conf-4.0.json .

In case they are not available at that URL or the unmarshalling fails for any reason the ones saved locally are used.

Would it be worth it removing the version from the file, so going to server-side-tls-conf.json ? ( since the version is in the file and the evaluation worker should always get the latest without recompiling )

### DIFF
--- a/worker/mozillaEvaluationWorker/configurations.go
+++ b/worker/mozillaEvaluationWorker/configurations.go
@@ -1,5 +1,6 @@
 package mozillaEvaluationWorker
 
+//saved TLS configurations
 var ServerSideTLSConfiguration = `{
     "configurations": {
         "modern": {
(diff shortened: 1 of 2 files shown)


0xdiba:
Also, if we want this URL ( or any configuration passed to workers ) to be configurable , we can maybe create a worker configuration file. To be content agnostic we can read it as a map[string] string and pass it to all the workers so every worker can get the values it knows and wants.

jvehent:
This looks pretty good, I'll merge it and we can improve on it later, specifically if we want a `latest` url.

PR #116, opened 2016-05-04
Remove stats.tc, which has been decom'd

claudijd:


claudijd:
Unsure why this build failed.  Checked travis-ci log, seemed unrelated to the change.

jvehent:
Builds fails when coming from external repos because travisci can't access the secrets. It's a security thing.

PR #124, opened 2016-06-30
CORS

0xdiba:
@jvehent r?

### DIFF
--- a/tlsobs-api/handlers.go
+++ b/tlsobs-api/handlers.go
@@ -125,7 +125,6 @@ func ScanHandler(w http.ResponseWriter, r *http.Request) {
 		err = errors.New("Could not process the requested scan")
 		return
 	}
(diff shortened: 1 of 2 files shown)


jvehent:
I'm not a fan, to be honest. A new router seems overkill when we just need to move the `setResponseHeader()` call further up so it works with more responses. The extra case for handling preflight requests in the cors routers add a new route path we should be handling in the multiplexer directly.

jvehent:
I think https://github.com/mozilla/tls-observatory/commit/9a8a92bbed439f42b82fda678e1d8967e3a96c31 should do it.

0xdiba:
No problem. 9a8a92b looks good.
I just thought that a new simple router would take this "complexity" out of the handler's logic.
Closing this one.

PR #125, opened 2016-07-15
TLS grading

0xdiba:
cc @jvehent
Not sure if we are good to merge yet.
The grading is working fine, but the Key exchange grading part is not yet identical to the SSLLabs one ( really close though ).

### DIFF
--- a/tools/extractCiphersuites.go
+++ b/tools/extractCiphersuites.go
@@ -25,7 +25,7 @@ type Encryption struct {
 	Bits   int    `json:"key"`
 }
 
-var opensslpath = "openssl"
+var opensslpath = "/opt/cipherscan/openssl"
 
 func main() {
 	cmd := opensslpath + " ciphers -v"
@@ -47,14 +47,23 @@ func main() {
 		if l == "" {
 			break
 		}
-		c := Cipher{}
 		line := strings.Fields(l)
-		c.Proto = line[1]
-		c.Kx = strings.Split(line[2], "=")[1]
-		c.Au = strings.Split(line[3], "=")[1]
-		c.Enc.Cipher = strings.Split(strings.Split(line[4], "=")[1], "(")[0]
-		c.Enc.Bits, _ = strconv.Atoi(strings.TrimRight(strings.Split(strings.Split(line[4], "=")[1], "(")[1], ")"))
-		c.Mac = strings.Split(line[5], "=")[1]
+		encbits, err := strconv.Atoi(strings.TrimRight(strings.Split(strings.Split(line[4], "=")[1], "(")[1], ")"))
+		if err != nil {
+			fmt.Errorf("Could not get encryption bits, %s", err)
+			return
+		}
+		enc := Encryption{
+			Cipher: strings.Split(strings.Split(line[4], "=")[1], "(")[0],
+			Bits:   encbits,
+		}
+		c := Cipher{
+			Proto: line[1],
+			Kx:    strings.Split(line[2], "=")[1],
+			Au:    strings.Split(line[3], "=")[1],
+			Enc:   enc,
+			Mac:   strings.Split(line[5], "=")[1],
+		}
 		ciphers[line[0]] = c
 	}
 
--- a/tools/extractCiphersuites.go
+++ b/tools/extractCiphersuites.go
@@ -1,0 +1,68 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os/exec"
+	"strconv"
+	"strings"
+)
+
+type Cipher struct {
+	// Name  string     `json:"name"`
+	Proto string     `json:"proto"`
+	Kx    string     `json:"kx"`
+	Au    string     `json:"au"`
+	Enc   Encryption `json:"encryption"`
+	Mac   string     `json:"mac"`
+}
+
+type Encryption struct {
+	Cipher string `json:"cipher"`
+	Bits   int    `json:"key"`
+}
+
+var opensslpath = "openssl"
+
+func main() {
+	cmd := opensslpath + " ciphers -v"
+	comm := exec.Command("bash", "-c", cmd)
+	var out bytes.Buffer
+	var stderr bytes.Buffer
+	comm.Stdout = &out
+	comm.Stderr = &stderr
+	err := comm.Run()
+	if err != nil {
+		log.Println(err)
+		return
+	}
+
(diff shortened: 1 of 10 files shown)


worker/mozillaGradingWorker/ciphersuites.go
```diff
@@ -0,0 +1,1295 @@
+package mozillaGradingWorker
+
+//OpenSSLCiphersuites contains all the ciphersuites recognised by OpenSSL
+var OpenSSLCiphersuites = `{
```
jvehent: I'm always concerned about maintainability when we define constants like these. I would prefer if we had a programmatic way to generate this json file, which turns out to be pretty easy with a few lines of bash.
```bash
#!/usr/bin/env bash
echo "{"
ctr=0
./cipherscan/openssl ciphers -V | while read code ign name proto kx au enc mac
do
    if [ $ctr -gt 0 ]; then echo ','; fi
    echo -n "\"$name\":{
    \"proto\": \"$proto\",
    \"kx\": \"$(echo $kx|cut -d '=' -f 2)\",
    \"au\": \"$(echo $au|cut -d '=' -f 2)\",
    \"mac\": \"$(echo $mac|cut -d '=' -f 2)\",
    \"encryption\": {
        \"cipher\": \"$(echo $enc | grep -Po 'Enc=\K[a-zA-Z0-9]+')\",
        \"key\": $(echo $enc | grep -Po 'Enc=[a-zA-Z0-9]+\(\K[0-9]+')
    }
  }"
    ctr=1
done
echo "
}"
```
I'm thinking that invoking a go generate function on build to create this file would be better than managing it by hand.

jvehent:
Besides one big comment about the way the openssl ciphers are generated, I think this looks pretty good.

What's missing is the integration with the tlsobs client, which requires implementing an AnalysisPrinter interface.

It would also be nice to have an Assertor to check stuff in the runner, but that's lower priority.

+:100: for getting this done!

worker/mozillaGradingWorker/ciphersuites.go
0xdiba: I generated the file by running the OpenSSL output through Go, so it is even more silly that I did not include it :no_mouth:

jvehent: Excellent! Then let's make it a go generate task that runs when we compile
the scanner so it's always up to date.

jvehent:
As a side note, it's worth noting this grading implements only small parts of the ssllabs grading. I think it's fine, and we will gradually improve it, but let's not assume that we have the quality level SSLLabs spent years of perfecting... yet.

marumari:
Integration with the client would certainly be nice, but for now all I need is for it to work on the API.  Woohoo!

0xdiba:
go:generate works. I doubt that it's textbook go:generate usage though :smiling_imp: 
@jvehent I have kept the original file ( in case the command runs into problems ). 
Do you prefer remove the file, add the go:generate command to the scanner docker-file and remove the original file?

tools/extractCiphersuites.go
```diff
@@ -0,0 +1,68 @@
...
+	Enc   Encryption `json:"encryption"`
+	Mac   string     `json:"mac"`
+}
+
+type Encryption struct {
+	Cipher string `json:"cipher"`
+	Bits   int    `json:"key"`
+}
+
+var opensslpath = "openssl"
```
jvehent: I think we should use the one in cipherscan rather than the system one. It will support more ciphers.

tools/extractCiphersuites.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		c := Cipher{}
+		line := strings.Fields(l)
+		c.Proto = line[1]
+		c.Kx = strings.Split(line[2], "=")[1]
+		c.Au = strings.Split(line[3], "=")[1]
+		c.Enc.Cipher = strings.Split(strings.Split(line[4], "=")[1], "(")[0]
+		c.Enc.Bits, _ = strconv.Atoi(strings.TrimRight(strings.Split(strings.Split(line[4], "=")[1], "(")[1], ")"))
+		c.Mac = strings.Split(line[5], "=")[1]
+		ciphers[line[0]] = c
```
jvehent: how about
```go
c := Cipher{
    Proto: line[1],
    Kx: strings.Split(line[2], "=")[1],
    Au: strings.Split(line[3], "=")[1],
    ...
}
```

jvehent: Meh, scratch zap, I'm nitpicking.

jvehent:
You should keep the original and replace its value with the output of go generate.

0xdiba:
We should be good to go now :beer:

jvehent:
Deployed in stage and works as expected: http://apistage201512031437-2sjrnmheum.elasticbeanstalk.com/api/v1/results?id=1237

Except for the tlsobs client which has two minor issues:
1) The grading working isn't imported:
```diff
diff --git a/tlsobs/main.go b/tlsobs/main.go
index 1f133bc..b7436e3 100644
--- a/tlsobs/main.go
+++ b/tlsobs/main.go
@@ -17,6 +17,7 @@ import (
        "github.com/mozilla/tls-observatory/database"
        "github.com/mozilla/tls-observatory/worker"
        _ "github.com/mozilla/tls-observatory/worker/mozillaEvaluationWorker"
+       _ "github.com/mozilla/tls-observatory/worker/mozillaGradingWorker"
 )
 
 func usage() {
```

2) The letter grade isn't displayed, only the numerical grade. 
```diff
diff --git a/worker/mozillaGradingWorker/mozillaGradingWorker.go b/worker/mozillaGradingWorker/mozillaGradingWorker.go
index 0ac43f8..e89feea 100644
--- a/worker/mozillaGradingWorker/mozillaGradingWorker.go
+++ b/worker/mozillaGradingWorker/mozillaGradingWorker.go
@@ -122,7 +122,8 @@ func (e eval) AnalysisPrinter(r []byte, targetLevel interface{}) (results []stri
                err = fmt.Errorf("Mozilla grading worker: failed to parse results: %v", err)
                return
        }
-       results = append(results, fmt.Sprintf("* Grade: %f", eval.Grade))
+       results = append(results, fmt.Sprintf("* Grade: %s (%.0f/100)",
+               eval.LetterGrade, eval.Grade))
        for _, e := range eval.Failures {
                results = append(results, fmt.Sprintf("  - %s", e))
        }
```

I added commits for those two. If they are OK, I'll deploy production.

0xdiba:
My bad!
Probably committed and then ran tlsobs :sweat_smile:

PR #126, opened 2016-07-17
tivial spelling fixes

ka7:


jvehent:
Good catch! Thanks for the patch.

PR #135, opened 2016-08-05
Display truststore validation info of certificate

jvehent:
This patchs adds check marks to indicate which truststores a certificate is valid for.
r? @0xdiba 

val info


jvehent:
Will submit this patch alongside another PR.

PR #136, opened 2016-08-05
Fixes to trust handling

jvehent:
r? @0xdiba


0xdiba:
Looks Good!
Nice touch with the truststore validation checkmarks.

I will create a PR with the error column we were talking about.
Merging?

Issue #139, opened 2016-09-02
Could not find db in request context


briner:
Hi,

I try to install tls-observatory on a Debian Server. I do have installed a postgresql server, I've followed what your [aws_create_env.sh](https://github.com/mozilla/tls-observatory/blob/master/tools/aws_create_env.sh) to create the postgres users and db.

Now, I got
```bash
# client
curl --data "target=www.unige.ch&rescan=False" http://127.0.0.1:8083/api/v1/scan
Could not access database.

# on server
./tlsobs-api-20160831+4b52dff.dev -c /etc/tls-observatory/api.cfg -debug
5time="2016-09-02T09:14:06+02:00" level=debug msg="Scan endpoint received request" form values=map[] headers=map[User-Agent:[curl/7.50.1] Accept:[*/*] Content-Length:[32] Content-Type:[application/x-www-form-urlencoded]] 
time="2016-09-02T09:14:06+02:00" level=error msg="Could not find db in request context" 
```

The ```-debug``` does not provide many information. How can I do to get the context ?

0xdiba:
Really strange error...
It indicates that the adapter part of [this](https://github.com/mozilla/tls-observatory/blob/master/tlsobs-api/main.go#L59) has not run.
It does not point to a db configuration issue ( the point where the database is accessed is never reached...).

Oddly enough, I cannot reproduce with debian jessie.

Can you use `go get github.com/mozilla/tls-observatory/tlsobs-api` and try to reproduce?

briner:
Hi,

In a new environment, I've done :
```bash
mkdir /home/briner/go
export GOPATH="/home/briner/go"
go get github.com/mozilla/tls-observatory/tlsobs-api
```

Then I run this fresh version of tls-api :
```bash
cd /home/briner/go/bin
./tlsobs-api -debug
  # time="2016-09-02T11:00:26+0./tlsobs-api -debug
time="2016-09-02T11:00:26+02:00" level=debug msg="Scan endpoint received request" form values=map[] headers=map[Content-Length:[32] Content-Type:[application/x-www-form-urlencoded] User-Agent:[curl/7.50.1] Accept:[*/*]] 
time="2016-09-02T11:00:26+02:00" level=error msg="Could not find db in request context" 
2:00" level=debug msg="Scan endpoint received request" form values=map[] headers=map[Content-Length:[32] Content-Type:[application/x-www-form-urlencoded] User-Agent:[curl/7.50.1] Accept:[*/*]] 
[...]

PR #140, opened 2016-09-08
Use stdlib context

0xdiba:
The addition of context to the stdlib causes problems with the context provided by gorilla framework.
#139 is caused by this. 
To merge this, the project will have to be Go1.7+ only. If we want to remain compatible with past version of go we can investigate other solutions.

@jvehent ?


jvehent:
I'm in favor. Maybe add a note to the README?

0xdiba:
Definitely.
Would it make sense to tag the most recent backwards compatible version?
Not that it would make much difference since upgrading is, in most cases,
trivial.

jvehent:
Nah. No need to maintain compatibility. People should use 1.7.

0xdiba:
@jvehent Merging?

PR #142, opened 2016-09-14
Add API endpoint to POST certificates

jvehent:
This is a large patch to add a new endpoint that accepts certificates submitted directly through the API. Most of the code is fairly straighforward, up until we get to the path building code. You see, the API does not have loaded truststores, and it seems silly to load them from local files when the database already has all the information needed. So I implemented path building using data in the certificate table. It's rather crude, and could probably be improved, but it gets the job done and has the benefit of checking for certificates signatures, so we know there isn't any false positive in the path.

@0xdiba, would you do me the honors of a review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Mozilla TLS Observatory
 
+Want the WebUI? Check out [Mozilla's Observatory](https://observatory.mozilla.org) !
+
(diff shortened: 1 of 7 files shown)


0xdiba:
LGTM :+1: 
Should we add a sample API call for this endpoint to the README?

PR #146, opened 2016-09-22
Add Subject+SPKI hashing to certificate parsing

jvehent:
This patch adds support for computing a hash of the raw subject and spki fields of a certificate and storing it in database. The new field is called `sha256_subject_spki` in both the DB and JSON output. It represents the hexadecimal string of the SHA256 of DER encoded Subject and SPKI fields.

Before deploying this to prod, we'll need to run this query: `ALTER TABLE certificates ADD COLUMN sha256_subject_spki;`

Then run the tool to update existing rows: `go run fixSHA256SubjectSPKI.go`

And finally set the `NOT NULL` condition on the column: `ALTER TABLE certificates ALTER COLUMN sha256_subject_spki SET NOT NULL;`

r? @mozkeeler
f? @sleevi (and thanks for the code snippet)

### DIFF
--- a/database/schema.sql
+++ b/database/schema.sql
@@ -2,6 +2,7 @@ CREATE TABLE certificates(
 	id	        				serial primary key,
 	sha1_fingerprint	        varchar NOT NULL,
 	sha256_fingerprint	    	varchar NOT NULL,
+    sha256_subject_spki         varchar NOT NULL,
 	pkp_sha256			    	varchar NOT NULL,
 	serial_number				varchar NULL,
 	version						integer NULL,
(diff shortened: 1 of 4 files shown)


database/schema.sql
```diff
@@ -2,6 +2,7 @@ CREATE TABLE certificates(
 	id	        				serial primary key,
 	sha1_fingerprint	        varchar NOT NULL,
 	sha256_fingerprint	    	varchar NOT NULL,
+    sha256_subject_spki         varchar NOT NULL,
```
mozkeeler: Looks like there's a tabs/spacees issue here (unless it's just a display issue in github...)

jvehent: Mixing tabs and spaces is never a good idea. I moved the entire file to spaces only.

PR #147, opened 2016-09-22
Use parameterized queries instead of Sprintf

jvehent:
blocked by #146 
r? @0xdiba

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,51 +12,31 @@ BUILDREV	:= $(BUILDDATE)+$(BUILDREF).$(BUILDENV)
 OS			:= linux
 ARCH		:= amd64
 
-ifeq ($(ARCH),amd64)
-	FPMARCH := x86_64
-endif
-ifeq ($(ARCH),386)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -29,7 +29,7 @@ BINDIR		:= bin/$(OS)/$(ARCH)
 GCC			:= gcc
 CFLAGS		:=
 LDFLAGS		:=
-GOOPTS		:= -tags netgo
+GOOPTS		:=
 GO 			:= GOOS=$(OS) GOARCH=$(ARCH) GO15VENDOREXPERIMENT=1 go
```
0xdiba: Since we change the file and we are Go 1.7+ only now, I think that `GO15VENDOREXPERIMENT=1` should go too.

0xdiba:
Other than that govendor nitpickism :smiling_imp:
:+1:

Makefile
jvehent: Good call, I fixed more Makefile stuff in the next commit.

PR #149, opened 2016-09-28
Add and remove entries from runner configuration

jasonthomas:
I've removed APK endpoints as they are no longer valid. I've added another everything.me and Kinto endpoint. Also cleaned up a bunch of white space chars (thanks vim).

r?


jvehent:
Nice! Thanks @jasonthomas

PR #160, opened 2016-10-24
PostCertificateHandler is not able to handle empty filenames

0xdiba:
change the filename to the hash of the certificate, that is always non-empty.


jvehent:
Good catch! Thanks.

PR #167, opened 2016-11-11
Add ev-checker worker

autrilla:
It's still missing API docs, but I'd like a review before I write those.


worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -0,0 +1,100 @@
...
+	var oid string
+	for _, k := range params.MapKeys() {
+		if k.Interface() == "oid" {
+			var ok bool
+			oid, ok = params.MapIndex(k).Interface().(string)
+			if !ok {
+				w.error("%s", fmt.Errorf("Could not cast oid to string"), res)
+			}
+		}
+	}
```
jvehent: Rather than reflecting by yourself, an inelegant solution that I used a couple times is to json marshal the interface, then unmarshal it into a local struct, like this:
```go
type myParamsStruct struct {
    OID string `json:"oid"`
}

paramsData, err := json.Marshal(in.Params)
var myparams myParamsStruct
err = json.Unmarshal(params.Data, &myparams)
```
The benefit of this approach is its generic to all workers, as opposed to writing custom reflection code everywhere. As such, it can be moved into the `workers` package instead of living here. For an example, take a look at https://github.com/mozilla-services/userplex/blob/master/modules/modules.go#L72-L80

worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -0,0 +1,100 @@
...
+	fmt.Println("oid:", oid)
```
jvehent: workers can, and probably should, have log message, so feel free to keep this, just make it nicer ;)

worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -0,0 +1,100 @@
...
+	certs, err := in.DBHandle.GetAllCertsInChain(in.Certificate.ID)
+	if err != nil {
+		w.error("Could not get all certificates in chain: %s", err, res)
+		return
+	}
+	var buffer bytes.Buffer
+	for i := len(certs) - 1; i >= 0; i-- {
```
jvehent: I bumped on this wondering why you didn't just use `for i, cert := range certs`. Maybe add a comment indicating why we're processing the certs from last to first?

worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -0,0 +1,100 @@
...
+		cert, err := certs[i].ToX509()
+		if err != nil {
+			w.error("Could not convert certificate to X509: %s", err, res)
+			return
+		}
+		err = pem.Encode(&buffer, &pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
+		if err != nil {
+			w.error("Error PEM-encoding certificate", err, res)
+			return
+		}
```
jvehent: `certificate.Raw` already contains the PEM, only it's a single string and not wrapped at 64 char. Rather than decoding the ASN.1 and reencoding the PEM, it would be nicer to just have `certificate.ToPEM` in the certificates package that returns the properly formatted PEM block.

worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	file, err := ioutil.TempFile("", "")
```
jvehent: add a `defer os.Remove(file)`

worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -0,0 +1,100 @@
...
+		w.error("Could not create temporary file to write certificates: %s", err, res)
+		return
+	}
+	defer file.Close()
+	_, err = buffer.WriteTo(file)
+	if err != nil {
+		w.error("Could not write certificates to temporary file: %s", err, res)
+		return
+	}
+	fmt.Println("ev-checker", "-o", oid, "-h", scan.Target, "-c", file.Name())
```
jvehent: to be removed?

worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -0,0 +1,100 @@
...
+	_, err = cmd.Output()
+	if err != nil {
+		w.error("Could not get output from ev-checker: %s", err, res)
+		return
+	}
+	res <- worker.Result{
+		Success:    true,
+		WorkerName: workerName,
+		Errors:     nil,
+		Result:     []byte(`"Success"`),
```
jvehent: @mozkeeler: Is there anything else you would like `ev-checker` to return? This is an `[]byte`, so technically it could contain anything, even the base64 of the code snippet you normally include in firefox.

worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -0,0 +1,100 @@
...
+	_, err = buffer.WriteTo(file)
+	if err != nil {
+		w.error("Could not write certificates to temporary file: %s", err, res)
+		return
+	}
+	fmt.Println("ev-checker", "-o", oid, "-h", scan.Target, "-c", file.Name())
+	cmd := exec.Command("ev-checker", "-o", oid, "-h", scan.Target, "-c", file.Name())
+	_, err = cmd.Output()
+	if err != nil {
+		w.error("Could not get output from ev-checker: %s", err, res)
```
jvehent: This is a nitpick, but `res` should probably be the first parameter, to avoid confusing it with a value of `Printf`

database/database.go
```diff
@@ -16,20 +16,21 @@ type DB struct {
...
+	Cert_id          int64                  `json:"cert_id"`
+	Trust_id         int64                  `json:"trust_id"`
+	Is_valid         bool                   `json:"is_valid"`
+	Validation_error string                 `json:"validation_error,omitempty"`
+	Complperc        int                    `json:"completion_perc"`
+	Conn_info        connection.Stored      `json:"connection_info"`
+	AnalysisResults  []Analysis             `json:"analysis,omitempty"`
+	Ack              bool                   `json:"ack"`
+	Attempts         int                    `json:"attempts"` //number of retries
+	AnalysisParams   map[string]interface{} `json:"analysis_params"`
```
jvehent: This is a nitpick, and I'm fine leaving this as a `map[string]interface{}`, but for the sake of the discussion I generally try to avoid `map[string]interface{}` as much as possible. `AnalysisResults`, for example, uses an `[]Analysis` where the name of the analysis worker is included in the `Analysis` struct.
```go
type Analysis struct {
        ID       int64           `json:"id"`
        Analyzer string          `json:"analyzer"`
        Result   json.RawMessage `json:"result"`
}
```
The upside of this approach is the analyzer name is typed. Downside: you need to iterate through the array to find the entry that applies to a given worker, instead of doing a hash lookup. In this case, the performance cost is irrelevant given we'll never have more than a couple entries in the array.

database/certificate.go
```diff
@@ -665,3 +665,56 @@ func (db *DB) IsTrustValid(id int64) (bool, error) {
...
 	return isValid, err
 }
+
+// GetAllCertsInChain returns the certificate chain of which tipCertID is the tip of,
+// which is to say it returns the certificate that trusts tipCertID, and the
+// certificate that trusts that other certificate, etc.
+func (db *DB) GetAllCertsInChain(tipCertID int64) ([]certificate.Certificate, error) {
+	rows, err := db.Query(`
+WITH RECURSIVE parents AS (
+	SELECT * FROM trust WHERE cert_id=$1
```
autrilla: I'm guessing it will, because I'm not sure what happens when you call `Scan` and get more rows than expected. We should probably also store all those columns in a variable somewhere instead of repeating them in every query, because we use the exact same fields in several functions. In fact, the scanCert function expects those fields.

tlsobs-scanner/main.go
```diff
@@ -234,7 +233,8 @@ func scan(scanID int64, cipherscan string) {
 	// launch workers that evaluate the results
 	resChan := make(chan worker.Result)
 	totalWorkers := 0
-	for _, wrkInfo := range worker.AvailableWorkers {
+	for k, wrkInfo := range worker.AvailableWorkers {
+		workerInput.Params, _ = scan.AnalysisParams[k]
```
autrilla: No, it will just pass `nil` to the worker. I thought it was a better idea for the worker to validate params, because they might be optional to begin with.

worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -0,0 +1,100 @@
...
+	"reflect"
+)
+
+var workerName = "evCheckerWorker"
+var workerDesc = `Provides insights for EV certificates`
+
+var log = logger.GetLogger()
+
+func init() {
+	worker.RegisterWorker(workerName, worker.Info{Runner: new(evWorker), Description: workerDesc})
```
autrilla: Maybe running `which ev-checker` to see if it's in `PATH`?

worker/evCheckerWorker/evCheckerWorker.go
autrilla: Err, this is just a leftover debugging message I missed when looking over the diff :p

jvehent:
A few more things:
1. add the new columns to `database/schema.sql`
2. add compilation of `ev-checker` to travisci and add it to the `tls-observatory-scanner` docker container

worker/evCheckerWorker/evCheckerWorker.go
jvehent: `os.Lstat` will do the job: https://golang.org/pkg/os/#Lstat

autrilla: I'm not sure how to cleanly pass options to workers, considering they're on a different package and we initialize workers with `init()`. We could have it as an exported variable on the worker, but then `init()` will always execute before we get a chance to change the variable, and we can't really have the worker read it from main because then we'd have an import loop.

conf/scanner.cfg
```diff
@@ -1,38 +0,0 @@
-[General]
```
jvehent: Why did you delete this file?

database/certificate.go
```diff
@@ -703,8 +638,20 @@ SELECT
...
-FROM certificates INNER JOIN (SELECT $1 UNION ALL SELECT issuer_id FROM parents) p(cert_id) ON p.cert_id=certificates.id`, tipCertID)
+	raw_cert`
+
+// GetAllCertsInChain returns the certificate chain of which endEntityID is the tip of,
+// which is to say it returns the certificate that signs endEntityID, and the
+// certificate that trusts that other certificate, etc.
+func (db *DB) GetAllCertsInChain(endEntityID int64) ([]certificate.Certificate, error) {
+	rows, err := db.Query(`
+WITH RECURSIVE parents AS (
+	SELECT ` + certRows + ` FROM trust WHERE cert_id=$1
```
jvehent: that doesn't make sense. you're selecting rows from the certificates table while querying the trust table.

database/certificate.go
```diff
@@ -703,8 +638,20 @@ SELECT
...
+	UNION ALL
+	SELECT trust.* FROM trust JOIN parents ON trust.cert_id=parents.issuer_id WHERE trust.cert_id IS DISTINCT FROM trust.issuer_id
```
jvehent: `trust.*` should be replaced with columns

database/certificate.go
```diff
@@ -666,18 +612,7 @@ func (db *DB) IsTrustValid(id int64) (bool, error) {
...
-func (db *DB) GetAllCertsInChain(tipCertID int64) ([]certificate.Certificate, error) {
-	rows, err := db.Query(`
-WITH RECURSIVE parents AS (
-	SELECT * FROM trust WHERE cert_id=$1
-	UNION ALL
-	SELECT trust.* FROM trust JOIN parents ON trust.cert_id=parents.issuer_id WHERE trust.cert_id IS DISTINCT FROM trust.issuer_id
-)
-SELECT                        
-	id,
+const certRows = `id,
```
jvehent: you mean `allCertificateColumns` ?

worker/evCheckerWorker/evCheckerWorker.go
```diff
@@ -9,29 +9,36 @@ import (
...
 )
 
 var workerName = "evCheckerWorker"
-var workerDesc = `Provides insights for EV certificates`
+var workerDesc = `Determines if a given EV policy fulfills the requirements of Mozilla's Root CA program.`
 
 var log = logger.GetLogger()
 
 func init() {
+	if _, err := os.Lstat(EvCheckerBinaryName); err != nil {
```
jvehent: I don't believe lstat will resolve the path. it just takes the absolute path to an inode.

autrilla: Sure, but I thought you said we should just take the location of the binary as a config option

database/certificate.go
autrilla: Actually, I think the query was fine as I wrote it the first time. The recursive part of the query just returns a bunch of trust entries, and then the select just gets all certificates with the ids in the trust entries.

conf/scanner.cfg
autrilla: Mistake, I meant not to check it in, it looks like I accidentally deleted it.

jvehent:
Per yesterday's discussion, try to build `evchecker` in travisci and only copy the resulting binary into the tlsobs scanner container, to avoid installing build tools in the container. If that doesn't work, the current strategy is still ok.

Note: you changes to `database/schema.sql` are still missing.

autrilla:
Not sure what changes to schema.sql you're referring to. I added the analysis_params column here: https://github.com/mozilla/tls-observatory/pull/167/files#diff-88655de7e12d7fa28ebbde6309bc75c2R75

jvehent:
> I added the analysis_params column here

Ha, sorry, I had missed it.

jvehent:
If I remember our conversation correctly, the remaining issue was changing the way post parameters are handled to be backward compatible with the current logic (specifically: take `parameters` from a form value, not from the entire post body).

jvehent:
I ran into a few issues in functional testing.

1. First, when running a normal scan without parameters, the scanner logs the following error:
`time="2016-11-18T08:57:04-05:00" level=error msg="Worker returned with errors" errors=[Invalid parameters passed to evWorker: <invalid reflect.Value>] worker_name=evCheckerWorker `
I would recommend taking the approach of unmarshaling and remarshaling as json, as posted in a previous comment, instead of using reflection directly.

2. Then, I tried passing parameters to the scan, as follows:
```bash
$ curl -X POST -d 'params={"evCheckerWorker":{"oid":"2.16.840.1.114412.2.1"}}' 'http://localhost:8083/api/v1/scan?target=addons.mozilla.org&rescan=true'
```
and got the following error: `time="2016-11-18T09:09:41-05:00" level=error msg="Worker returned with errors" errors=[Could not get output from ev-checker: exit status 1] worker_name=evCheckerWorker`
There isn't enough information in the logs to figure out why it fails. I think increasing log details is important before we merge this. (log both parameters and command executed, like we do for cipherscan).

Additionally:
3. Maybe change the name of the worker. `evCheckerWorker` should just be `ev-checker` or `evChecker`.
4. Document the new parameters in the API doc in the README.

worker/evCheckerWorker/evCheckerWorker.go
jvehent: @autrilla : After giving this some thoughts, I think we should return the `cmd.Output()` in `Result`. That way, if @mozkeeler decides to add more details to what `ev-checker` prints, we won't need to change anything on our side to handle it.

static/evchecker.html
```diff
@@ -0,0 +1,93 @@
...
+  </form>
+  <p id="result"/>
+  <script>
+function checkResult(id) {
+    return fetch("/api/v1/results?id=" + id)
+	.then(function(response) {
+            return response.json().then(function(json) {
+		let completion_percent = json.completion_perc;
+		if (completion_percent !== 100) {
+		    console.log("Not done yet, running again");
```
jvehent: Also print a dot in the html to indicate to give visual feedback to the user that the scan is progressing. The command line tool does this.

static/evchecker.html
```diff
@@ -0,0 +1,93 @@
+  <style>
```
jvehent: nit: this should be called `ev-checker.html` for consistency.

static/evchecker.html
```diff
@@ -0,0 +1,93 @@
...
+
+function send() {
+    var target = document.getElementById("target").value;
+    var oid = document.getElementById("oid").value;
+    var rootCertificate = document.getElementById("rootCertificate").value;
+    var result = document.getElementById("result");
+    startScan(target, oid, rootCertificate).then(function(id) {
+        console.log("Scan started with id", id);
+        result.innerHTML = "Scan started, waiting for result...";
+	return checkResult(id).then(function(params) {
```
jvehent: nit: missing space in indent

static/evchecker.html
```diff
@@ -0,0 +1,93 @@
...
+	    if (params !== undefined) {
+                result.innerHTML = "ev-checker reported success";
+                result.style.color = "Green";
+	    } else {
+                result.innerHTML = "ev-checker reported failure";
+                result.style.color = "Red";
```
jvehent: In case of failure, we need to return the stdout from ev-checker to the user. It looks like a non-zero exit code from ev-checker means the output isn't stored, so you'll need to fix that first.

static/evchecker.html
```diff
@@ -0,0 +1,93 @@
...
...led .pure-menu-link:hover{background-color:transparent}.pure-menu-active>.pure-menu-link,.pure-menu-link:hover,.pure-menu-link:focus{background-color:#eee}.pure-menu-selected .pure-menu-link,.pure-menu-selected .pure-menu-link:visited{color:#000}.pure-table{border-collapse:collapse;border-spacing:0;empty-cells:show;border:1px solid #cbcbcb}.pure-table caption{color:#000;font:italic 85%/1 arial,sans-serif;padding:1em 0;text-align:center}.pure-table td,.pure-table th{border-left:1px solid #cbcbcb;border-width:0 0 0 1px;font-size:inherit;margin:0;overflow:visible;padding:.5em 1em}.pure-table td:first-child,.pure-table th:first-child{border-left-width:0}.pure-table thead{background-color:#e0e0e0;color:#000;text-align:left;vertical-align:bottom}.pure-table td{background-color:transparent}.pure-table-odd td{background-color:#f2f2f2}.pure-table-striped tr:nth-child(2n-1) td{background-color:#f2f2f2}.pure-table-bordered td{border-bottom:1px solid #cbcbcb}.pure-table-bordered tbody>tr:last-child>td{border-bottom-width:0}.pure-table-horizontal td,.pure-table-horizontal th{border-width:0 0 1px;border-bottom:1px solid #cbcbcb}.pure-table-horizontal tbody>tr:last-child>td{border-bottom-width:0}
```
jvehent: Maybe the styling should be in a separate file? The license block above only applies to it, not to the code you wrote.

README.md
```diff
@@ -187,6 +187,7 @@ $ curl -X POST 'https://tls-observatory.services.mozilla.com/api/v1/scan?target=
 * `params` JSON object in which each key represents one of TLS Observatory's workers. The value under each key will be passed as the parameters to the corresponding worker. For example, `{"ev-checker": {"oid": "foo"}}` will pass `{"oid": "foo"}` to the ev-checker worker. The following workers accept parameters:
   * ev-checker: Expects a JSON object with the following keys:
     * oid: the oid of the EV policy to check
+    * rootCertificate: the root certificate to check against, in PEM format
```
jvehent: Since this is a bit tricky, a bash example of how to call this on the command line would be great.

jvehent:
Looking good. I can get the EV test on AMO to pass through the webui. I left a few comments but we're getting close to merging point 👍

static/evchecker.html
autrilla: As far as I can tell the result from failed workers isn't shown on the endpoint, so this would be hard to do.

static/evchecker.html
autrilla: We could just pull it from a CDN if you want, or just have the style as a separate static file. I just added some styling because it looked awful unstyled.

jvehent: I'd prefer a separate static file. Maybe we'll write other UIs in the future that will use it as well.
Also, I don't think it needs to be minified, we're not concerned about performance here.

PR #171, opened 2016-11-18
add docker-compose

0xdiba:
Add docker-compose.yml for easy end to end testing
Introduced postgres Dockerfile  for easy setup. DB Schema is created automatically.


jvehent:
Should we have a 4th container running a test app to scan against?

0xdiba:
We can just run against mozilla.org, google.com or even badssl.
Would having a simple test app locally help?

The only use case I can think about right now is running end to end integration test through CI.
In that case having a local test app would prove useful for sure.

jvehent:
One benefit of running locally is we could verify various confs are measured correctly. We can't do that when targeting an external site.

0xdiba:
Should the app be trusted by the observatory?
So maybe add a "test" truststore?

jvehent:
Maybe we don't need a trusted cert, just a TLS endpoint. That still requires some scripting, but less than adding a new truststore.

jvehent:
Very nice!
I haven't used docker compose before. How would we go about starting the service and running a scan agains that test server in travisci?

0xdiba:
I think travis supports docker-compose by default.
Just a `docker-compose up` will bring up all the services connected with each other.
The only part missing is the curl command performing the scan request against the api.
This probably has to be issued in the travis.yml. I will search around for a bit and commit it if that works.

tools/simple-server.go
```diff
@@ -0,0 +1,32 @@
...
+		PreferServerCipherSuites: true,
+		CipherSuites: []uint16{
+			tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+			tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
+			tls.TLS_RSA_WITH_AES_256_GCM_SHA384,
+			tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+		},
+	}
+	srv := &http.Server{
+		Addr:         ":443",
```
jvehent: Might want to pick a port higher than 1024 to avoid having to run as root.

jvehent:
Instead of the curl command, we could use the `tlsobs` client like this: `tlsobs -targetLevel modern localhost:8443`
It will exit with a non-zero error code, which will fail travisci, if the measured level isn't modern.

0xdiba:
Good call. And we already have it in a docker container right?

tools/simple-server.go
0xdiba: The scanner currently connect on port 443 by default.
For this to work we'll have to make the port configurable.

jvehent:
Yes, you can run it with `$ docker run mozilla/tls-observatory tlsobs <TARGET>`

tools/simple-server.go
jvehent: Ha, I forgot about that. I guess `:443` is fine then :)

0xdiba:
Is there a specific reason we run `make` inside the travis build?
Do we need the produced executables?

jvehent:
`make` is only run as a test to verify everything tests and compiles appropriately

jvehent:
from travisci:
> panic: Post http://localhost:8083/api/v1/scan?target=sampleapp: read tcp 127.0.0.1:57040->127.0.0.1:8083: read: connection reset by peer

It's getting there. I'd love to merge it and have better testing when ready!

0xdiba:
It's probably a timing error with the start of the services.
I will try to fix it in the next few days and finally merge it 😅 .

jvehent:
```
$ docker run --net=host mozilla/tls-observatory tlsobs --targetLevel intermediate --observatory http://localhost:8083 sampleapp

panic: Post http://localhost:8083/api/v1/scan?target=sampleapp: read tcp 127.0.0.1:44866->127.0.0.1:8083: read: connection reset by peer
```

jvehent:
@0xdiba : want to give this another pass?

0xdiba:
Yep, @jvehent I will take a ( hopefully final ) look this week. ⏳

jvehent:
Let's close this for now, it's not highly needed, and we can always revisit later on.

Issue #179, opened 2016-11-30
Improvements to the ev-checker webui


jvehent:
1. Top entry area (currently default text "Target")
  * Add label: "Website URL"
  * Add example text in the entry area: "https://observatory.mozilla.org/"
  * Add ability to succeed with and without "https://" being entered

2. Second entry area (currently default text "OID Policy")
  * Add label: "EV Policy OID"
  * Add example text in the entry area: "2.23.140.1.1"
  * Add ability to succeed with and without "https://" being entered

3. PEM entry area
  * Add label: "Certificate PEM"
  * Add example text in the entry area:
```
-----BEGIN CERTIFICATE-----
...
-----END CERTIFICATE-----
```
4. Add `Browse...` button with same functionality as the `Browse...`  button in https://people-mozilla.org/~dkeeler/certsplainer/ and have it populate the PEM.

5. For Success output, also include the data needed to be added to  ExtendedValidation.cpp. See: https://wiki.mozilla.org/PSM:EV_Testing_Easy_Version#Success

jvehent:
Those changes should be made to https://github.com/mozilla/tls-observatory/blob/master/static/ev-checker.html and https://github.com/mozilla/tls-observatory/blob/master/static/ev-checker.js

jvehent:
I took care of 1, 2 and 3.
@autrilla: do you have an elegant way to integrate 4 with your JS lib?
@mozkeeler: we discussed hidding 5 from the normal output but @WilsonKathleen would like it displayed in the webui. Since the tls observatory displayed the raw output from `ev-checker`, I'll let you two judge which changes are appropriate, and we'll simply rebuild `ev-checker` with it.

autrilla:
It should be fairly easy to add a button and using https://developer.mozilla.org/en-US/docs/Web/API/FileReader populate the PEM field. Something like this as the action of a "Select file" button (so `type="file"`):

```javascript
function(e) {
    var reader = new FileReader();
    reader.onload(function(event) {
        document.getElementById("rootCertificate").value = event.target.result;
    });
    reader.readAsText(e.target.files[0])
}
```

jvehent:
This has shipped, except for 5 that belongs to @mozkeeler. Closing this issue, follow-up, if needed, can happen in the ev-checker repo.

PR #181, opened 2016-12-01
Add parameters validation to ev-checker.js

jvehent:
Some changes related to #179: input validation in javascript with feedback in the webui.
r? @autrilla


static/ev-checker.js
```diff
@@ -71,3 +83,14 @@ function send(e) {
             result.style.color = "Red";
 	});
 }
+
+function hostname_from(target) {
+    // if target is a URI, extract hostname from it
+    if (target.startsWith("http") == true) {
```
autrilla: nit: you don't need the `== true`

Issue #129, opened 2016-07-18
Use SSLLabs's client hello database to determine oldest clients


april:
The mozillaEvaluationWorker is only returning level and failures; but looking at the code it seems like it should return a lot more, such as the oldest clients that are supported.

jvehent:
The oldest clients get printed through the `AnalysisPrinter` interface, which really just maps the measured level to the Server Side TLS Guidelines. https://github.com/mozilla/tls-observatory/blob/master/worker/mozillaEvaluationWorker/mozillaEvaluationWorker.go#L718

We could return the oldest clients in the API, or you could grab the JSON of the guidelines and do the same thing on your side. I'm open to returning it in the tlsobs api if you want, it's a simple addition.

jvehent:
When you say `should return a lot more`, is there anything else than oldest clients you have in mind?

marumari:
There just seem to be a lot of things in here:

https://github.com/mozilla/tls-observatory/blob/08fc009cd747be17b664955f4812990bd2058fa1/worker/mozillaEvaluationWorker/configurations.go

That don't show up through the API or anything.  The TLS thing seems like a nice list, but I end up having to loop through each cipher suite to get the list, etc.  Even just having oldest_clients would be really nice right.

Also it seems that the oldest_client output is based solely upon the various compatibility levels; there doesn't seem to be anything deeper.  I assume that's a WONTFIX, but for example something like pokeinthe.io returns nothing:

```
$ tlsobs pokeinthe.io
Scanning pokeinthe.io (id 12119744)
Retrieving cached results from 6h3m21.184641455s ago. To run a new scan, use '-r'.

--- Certificate ---
Subject  CN=pokeinthe.io
SubjectAlternativeName
- pokeinthe.io
- www.pokeinthe.io
Validity 2016-06-27T05:15:00Z to 2016-09-25T05:15:00Z
CA       false
SHA1     951D06C916A77FB0660A7C6C42804473BF2E2336
SHA256   A4EC18F605D374CBF77600F43AB573D72DDA55F7CC02F9FCBEBCFFCBC7AD23CD
SigAlg   SHA256WithRSA
Key      ECDSA 384bits P-384

--- Chain of trust ---
0:	CN=pokeinthe.io
	type: end entity
	key: ECDSA 384bits P-384
	pin-sha256: llJLhjXjqfdJrklTYSM+vWgaBa4F0PgbAMXe+VanIMc=

[...]

PR #182, opened 2016-12-03
Ssl labs client support

jvehent:
This new worker and surrounding changes implement #129 : a method to determine client support using SSLLabs publicly provided data. The analysis worker uses cipherscan output and compares each server ciphersuite with the ciphersuites supported by the client. If a match is found (using ciphersuites code points), the client is marked as supported and the matching ciphersuite is recorded.

r? @0xdiba and @marumari


jvehent:
I added a few commits to make this better.

First of all, we don't use generated code anymore. Instead, a new constants package has an almost exhaustive list of TLS parameters that any other package can use. It might also be fun to export those constants in API endpoints, so other people can rely on them.

Second, the client support analyser now check for PFS and protocol support, and stores that information in the output.

This is starting to look good. Here's an example output: https://gist.github.com/jvehent/e870eb135424205a658b21f5e71dfaf6

jvehent:
@marumari : I'm particularly interested in your take on the JSON schema of the analysis output, and if we should break up the product name (Firefox, Chrome, Safari) and its version (27, 28, 29) to make it easier to detect the oldest supported version for a given browser.

jvehent:
@marumari: do you have any opinion on the output format, or is this one to your liking?
https://gist.github.com/jvehent/e870eb135424205a658b21f5e71dfaf6

marumari:
Err, I swear I responded.  WTF...

marumari:
```
    "IE 8-10": {
      "is_supported": false
}
```

This is a bit awkward, I'd rather simply check them each and create the string myself.  This would probably be a lot easier if the format was, instead of this:

```json
    "Chrome 39": {
      "code": 49195,
      "curve": "secp256r1",
      "protocol": "TLSv1.2",
      "ciphersuite": "ECDHE-ECDSA-AES128-GCM-SHA256",
      "is_supported": true
    },
    "Chrome 40": {
      "code": 49195,
      "curve": "secp256r1",
      "protocol": "TLSv1.2",
      "ciphersuite": "ECDHE-ECDSA-AES128-GCM-SHA256",
      "is_supported": true
}
```

Something like this:

```json
Chrome: {
  "39": {
    "code": 49195,
    "curve": "secp256r1",
    "protocol": "TLSv1.2",
    "ciphersuite": "ECDHE-ECDSA-AES128-GCM-SHA256",
    "is_supported": true
  },
  "40": {
    "code": 49195,
    "curve": "secp256r1",
    "protocol": "TLSv1.2",
    "ciphersuite": "ECDHE-ECDSA-AES128-GCM-SHA256",
    "is_supported": true
  }
}
```

You could also do it like this:
Chrome: `[[39, {...}], [40, {...}]]`

They both have their advantages and disadvantages: the latter makes it easier to find the most recent version, the prior makes it easier to query a single version without iterating the list.

Also, instead of this:

```
  "curve": "secp256r1",
  "protocol": "TLSv1.2",
  "ciphersuite": "ECDHE-ECDSA-AES128-GCM-SHA256"
```

You may want to have code points instead, so you're not constantly repeating long strings.  Up to you.  :)

jvehent:
I added the curve and protocol codes (ciphersuite was already added).

Version will always have to be a string, not a float, because each product handles it differently. That makes finding the oldest version a little tricky. We'll need special code to handle that 😞 

The next best thing is to go with a format like the following:
```json
    {
      "id": 149,
      "analyzer": "sslLabsClientSupport",
      "result": [
        {
          "name": "Android",
          "curve": "secp256r1",
          "version": "4.4.2",
          "protocol": "TLSv1.2",
          "curve_code": 23,
          "ciphersuite": "ECDHE-ECDSA-AES256-GCM-SHA384",
          "is_supported": true,
          "protocol_code": 771,
          "ciphersuite_code": 49196
        },
        {
          "name": "Android",
          "curve": "secp256r1",
          "version": "5.0.0",
          "protocol": "TLSv1.2",
          "curve_code": 23,
          "ciphersuite": "ECDHE-ECDSA-AES128-GCM-SHA256",
          "is_supported": true,
          "protocol_code": 771,
          "ciphersuite_code": 49195
        }
      ]
    }
```

PR #194, opened 2017-01-19
Dockerfile best practice

rzmf:


jvehent:
Thanks. Could you wrap the lines for readability, such that each individual command is on its own line? Like this: https://github.com/mozilla-services/hindsight/blob/master/Dockerfile#L20-L62

Issue #193, opened 2017-01-19
any way to use a own copy of server-side-tls-conf.json?


rzmf:
In https://github.com/mozilla/tls-observatory/pull/102#issuecomment-184386474 this was mentioned but never coded, or did i missed this?

0xdiba:
For the [hosted service](http://tls-observatory.services.mozilla.com/) No, you cannot change it.

If you are compiling on your own you can change https://github.com/mozilla/tls-observatory/blob/master/worker/mozillaEvaluationWorker/mozillaEvaluationWorker.go#L31 to point to your own hosted file. 
You can even set the URL to something that fails ( probably any other valid or invalid URL 😃  ) and the system will automatically fall back to [this file](https://github.com/mozilla/tls-observatory/blob/master/worker/mozillaEvaluationWorker/configurations.go) which you can change at will.

rzmf:
I run my own copy of scanner/api/"cli" as docker images.

Which part of the three actually does the download?

0xdiba:
The download is performed by the scanner.

rzmf:
ok, thanks for info.
A new option in scanner.cfg and commandline would be nice

rzmf:
You are sure about the download by the scanner?

# docker run -ti --rm mozilla/tls-observatory
root@f56cf65011bd:/go# tlsobs
time="2017-01-19T11:53:07Z" level=error msg="Get https://statics.tls.security.mozilla.org/server-side-tls-conf.json: dial tcp 54.240.162.44:443: i/o timeout"
time="2017-01-19T11:53:07Z" level=error msg="Could not get tls confs from url - fallback to locally saved configurations"

note that no observatory  (as in reachable api) was given and that tlsobs does not reach public internet

jvehent:
We should fix that, the tlsobs client doesn't use the JSON file. The reason it downloads it is because the tlsobs client imports the mozillaEvaluationWorker to print the analysis results, and the worker downloads the JSON file in its init function. That's only used when the worker runs as part of the scanner, but because init() is always executed when a program start, the tlsobs client ends up downloading it as well.

0xdiba:
Oh. I didn't think of that.
We should probably move any setup code to either the first time `worker.Run` is executed or to a new function of the worker interface ( e.g. `setup` ) that gets called by the scanner for all registered workers.

rzmf:
it does work, but it slows down the start of the request of the scan from the api. If api&scanner also try to download the file.....

[...]

PR #195, opened 2017-01-20
Lazily setup evaluation worker

0xdiba:
Based on discussion for #193 .

### DIFF
--- a/worker/mozillaEvaluationWorker/mozillaEvaluationWorker.go
+++ b/worker/mozillaEvaluationWorker/mozillaEvaluationWorker.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/mozilla/tls-observatory/certificate"
 	"github.com/mozilla/tls-observatory/connection"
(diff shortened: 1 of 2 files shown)


worker/mozillaEvaluationWorker/mozillaEvaluationWorker.go
```diff
@@ -33,25 +34,32 @@ var sstlsURL = "https://statics.tls.security.mozilla.org/server-side-tls-conf.js
...
-		log.Error("Could not get tls confs from url - fallback to locally saved configurations")
+		log.Warn(err)
+		log.Warn("Could not get tls confs from url - fallback to locally saved configurations")
 		// Try to continue with the locally hosted TLS configurations
 		err = json.Unmarshal([]byte(ServerSideTLSConfiguration), &sstls)
 		if err != nil {
 			log.Error(err)
 			log.Error("Could not load Server Side TLS configuration. Evaluation Worker not available")
+			//if we reach this stage, no TLS configs are loaded, so there is no need to have this worker available.
+			worker.RemoveWorker(workerName)
```
jvehent: I think we should just panic here. I can't think of a situation where I'd want to scanner to continue when loading the local conf has failed.

0xdiba: I don't think this is a valid scenario either, since fallback configurations are embedded in the code.
I will keep the `RemoveWorker` function in the worker just for consistency.

PR #191, opened 2017-01-05
Minor cleanups (and XSS fix)

KellerFuchs:
- [x] Removes some minor repetition.
- [x] Fixes minor HTML syntax issues.
- [x] Replaces HTML constructed with raw strings with DOM manipulation.
      This fixes a few XSS issues (that were mitigated by the CSP anyway); it also ought to be faster (`innerHTML` causes the text to be reparsed), but that doesn't really matter here.

PR #196, opened 2017-01-22
Minor cleanup

g-k:
This is a rebased version of #191 with the patch from https://github.com/mozilla/tls-observatory/pull/191#issuecomment-274352475.

I ran through the functional tests in that comment as well.


KellerFuchs:
@g-k Awesome, thanks.

Issue #198, opened 2017-01-26
Expose statistics about issuers from the API


jvehent:
@mozmark is interested in calculating the impact of revoking a given root or intermediate in OneCRL. We have some of that data in the trust table of the observatory, and should expose it through an API endpoint.

For a given certificate, we first need to find its internal ID. This can be done by querying the `certificates` table using the SHA256 of the certificate.
```sql
observatory=> select id from certificates where sha256_fingerprint = upper('25847d668eb4f04fdd40b12b6b0740c567da7d024308eb6c2c96fe41d9de218d');
   id    
---------
 1707992
(1 row)
```
The `trust` table can then be queried to retrieve a count of certificates issued by this certificate.
```sql
observatory=> select count(distinct(cert_id)) from trust where issuer_id=1707992;
 count 
-------
 31473
(1 row)
```
This will need to be done recursively to cover roots issuing intermediates issuing intermediates issuing end entities.

In the API, we can either create a new handler as `GET /api/v1/issuerstats?id=<certid>` (that should also take a sha256 instead of the id), or we can add the information to the existing `GET /api/v1/certificate` endpoint. The later might be nicer but could slow down certificate retrieval for the common use case, so maybe hiding the stat functionality behind a query string boolean would be needed.

The new handler function should be added to https://github.com/mozilla/tls-observatory/blob/master/tlsobs-api/handlers.go with a new route defined in https://github.com/mozilla/tls-observatory/blob/master/tlsobs-api/router.go. The corresponding queries need to live in the database package at https://github.com/mozilla/tls-observatory/blob/master/database/certificate.go.

jvehent:
> For a given certificate, we first need to find its internal ID. This can be done by querying the certificates table using the SHA256 of the certificate.

Note that you can just use [database.GetCertIDBySHA256Fingerprint](https://godoc.org/github.com/mozilla/tls-observatory/database#DB.GetCertIDBySHA256Fingerprint).

PR #203, opened 2017-02-01
#198 - Expose statistics about issuers from the API

mozmark:


database/certificate.go
```diff
@@ -388,6 +388,38 @@ func (db *DB) GetAllCertsInStore(store string) (out []certificate.Certificate, e
 	}
 }
 
+func (db *DB) GetAllCertIDsIssuedBy(sha256 string)(out []int64, err error) {
```
jvehent: nit: the naming convention for DB function contains the hash type, so something like `GetAllCertIDsBySHA256Fingerprint`.

tlsobs-api/router.go
```diff
@@ -78,6 +78,12 @@ var routes = Routes{
 		"/api/v1/truststore",
 		TruststoreHandler,
 	},
+	Route{
+		"Truststore",
```
jvehent: This seems incorrect.

tlsobs-api/router.go
```diff
@@ -78,6 +78,12 @@ var routes = Routes{
...
+		"GET",
+		"/api/v1/issuerStatus",
```
jvehent: `Status` or `stats`?
I'd also prefer to avoid camel case in api endpoints, and use lowercase only.

database/certificate.go
jvehent: Two more comments:
1. add a comment string that describes what the method does
2. does this return only end-entities or intermediates as well? I think the naming of the function should make that as unambiguous as possible. Maybe `GetEndEntityCertsForIssuerBySha256Fingerprint`. It's a mouthful, but there would be no confusion as to what it does.

tlsobs-api/handlers.go
```diff
@@ -439,6 +439,38 @@ func TruststoreHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// IssuerStatusHandler handles the /issuerStatus endpoint of the api.
+// It queries the database for a list of end-entity cert_ids which chain via the
+// given certificate.
+// It takes the following HTTP parameter:
+//     fingerprint - a base64 encoded sha256 certificate fingerprint
```
jvehent: You mean hex encoded, not base64, right?

tlsobs-api/handlers.go
```diff
@@ -439,6 +439,38 @@ func TruststoreHandler(w http.ResponseWriter, r *http.Request) {
...
+func IssuerStatusHandler(w http.ResponseWriter, r *http.Request) {
+	val := r.Context().Value(ctxDBKey)
+	if val == nil {
+		httpError(w, r, http.StatusInternalServerError,
+			fmt.Sprintf("Could not find database handler in request context"))
+		return
+	}
+	db := val.(*pg.DB)
+	certIDs, err := db.GetAllCertIDsIssuedBy(r.FormValue("fingerprint"));
```
jvehent: On other endpoints, we use `sha256` instead of fingerprint. I'd prefer to stay consistent.

tlsobs-api/handlers.go
```diff
@@ -439,6 +439,38 @@ func TruststoreHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	certIDsJson, marshalErr := json.Marshal(certIDs)
+
+	if marshalErr != nil {
+	httpError(w, r, http.StatusInternalServerError,
+			fmt.Sprintf("Unable to marshal certificate IDs"))
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(certIDsJson)
```
jvehent: We'll be returning a *lot* of data for roots & intermediates that are widely used. Imagine getting the stats of the intermediate of letsencrypt, the list could contain hundreds of thousands of IDs. That's too much for a single API call.

Two options:
1. only return stats, like a count of EE certs
2. paginate, which is a bit trickier to implement

We could also add more filtering parameters to the endpoint to reduce the scope of the search. I think 500 IDs should be a reasonable upper limit.

database/certificate.go
```diff
@@ -388,7 +388,12 @@ func (db *DB) GetAllCertsInStore(store string) (out []certificate.Certificate, e
 	}
 }
 
-func (db *DB) GetAllCertIDsIssuedBy(sha256 string)(out []int64, err error) {
+// GetEndEntityCertsForIssuerBySha256Fingerprint gets the certificate IDs of all
+// end entity certificates in the database that chain to the certificate with
+// the specified Sha256 fingerprint.
+func (db *DB) GetEndEntityCountForIssuerBySha256Fingerprint(sha256 string)(id int64, err error) {
```
jvehent: The returned value should be `count`, not `id`.

tlsobs-api/handlers.go
```diff
@@ -439,35 +439,35 @@ func TruststoreHandler(w http.ResponseWriter, r *http.Request) {
...
 			fmt.Sprintf("Could not find database handler in request context"))
 		return
 	}
 	db := val.(*pg.DB)
-	certIDs, err := db.GetAllCertIDsIssuedBy(r.FormValue("fingerprint"));
+	count, err := db.GetEndEntityCountForIssuerBySha256Fingerprint(r.FormValue("sha256"));
 
 	if err != nil {
 	httpError(w, r, http.StatusInternalServerError,
 			fmt.Sprintf("Unable to obtain a certificate for the given hash"))
```
jvehent: Maybe `Unable to retrieve statistics for the given issuer` ?

tlsobs-api/router.go
```diff
@@ -110,6 +116,12 @@ var routes = Routes{
 		PreflightHandler,
 	},
 	Route{
+		"CORS Preflight",
+		"OPTIONS",
+		"/api/v1/issuerStatus",
```
jvehent: woops, this needs to be changed to `/api/v1/issuereecount`

PR #210, opened 2017-02-09
Prevent infinite loops from breaks certificate paths building

jvehent:
Cross-signing in the web pki apparently contains a lot of infinite loops
that break the building of paths. This patch implement a simple []string
that contains the CNs of certs that have already been seen in the path,
to prevent the function from looping indefinitely.

### DIFF
--- a/database/certificate.go
+++ b/database/certificate.go
@@ -600,16 +600,19 @@ func (db *DB) GetValidationMapForCert(certID int64) (map[string]certificate.Vali
 }
 
 // GetCertPaths returns the various certificates paths from the current cert to roots.
-// It takes a certificate as argument that will be used as the start of the path, and
-// a genealogy slice that contains the list of known parents, to prevent looping. When
-// first invoked, the genealogy variable should be an empty []string.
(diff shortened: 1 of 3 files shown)


database/certificate.go
```diff
@@ -599,12 +599,17 @@ func (db *DB) GetValidationMapForCert(certID int64) (map[string]certificate.Vali
 	return certificate.GetValidityMap(ubuntu, mozilla, microsoft, apple, android), issuerId, nil
 }
 
-func (db *DB) GetCertPaths(cert *certificate.Certificate) (paths certificate.Paths, err error) {
+// GetCertPaths returns the various certificates paths from the current cert to roots.
+// It takes a certificate as argument that will be used as the start of the path, and
+// a genealogy slice that contains the list of known parents, to prevent looping. When
+// first invoked, the genealogy variable should be an empty []string.
```
autrilla: nit: I think the public API should not include the `[]string`, since after all we're asking to pass an empty slice the first call. I think we should instead have a private recursive function and make the public function create the slice for the caller.

jvehent: fixed

PR #215, opened 2017-03-18
Certificate insert optimizations

autrilla:
```
Before the optimization: BenchmarkInsertNewCertificate-8   	     100	  16625509 ns/op
After the optimization:  BenchmarkInsertNewCertificate-8   	     100	  11918866 ns/op
```

So ~30% faster now.

There's still plenty to optimize, as there's another N+1 problem with the truststore validation checks. My guess is that will also add another 30% speed improvement.

### DIFF
--- a/tlsobs-api/insert_test.go
+++ b/tlsobs-api/insert_test.go
@@ -1,0 +1,134 @@
+package main
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"log"
+	"os"
+	"testing"
+
+	"github.com/mozilla/tls-observatory/certificate"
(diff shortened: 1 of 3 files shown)


tlsobs-api/insert_test.go
```diff
@@ -0,0 +1,134 @@
+package main
```
jvehent: Would this be best placed in a `benchmark` subdirectory of the database folder, since we shouldn't run it as part of the normal `go test` process?

autrilla:
The benchmarks won't run unless `-bench .` is passed to `go test`, but I'm
fine with putting this in another directory.

PR #220, opened 2017-05-03
remove browserid.org (and diresworb.org)

jrgm:
r? - @jvehent


floatingatoll:
cc @rfk

jrgm:
ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1325196#c12

PR #221, opened 2017-05-23
Update relud's projects in tls-observatory

relud:
persona and universalsearch are dead
pageshot is now screenshots
normandy has two more urls
embedly-proxy is tiles not testpilot

### DIFF
--- a/conf/runner.yaml
+++ b/conf/runner.yaml
@@ -81,22 +81,8 @@ runs:
 
         # Normandy
         - normandy.cdn.mozilla.net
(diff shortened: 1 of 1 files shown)


relud:
also, I'm pretty sure the intermediate level monitoring should include normandy. is that for HPKP?

jvehent:
> also, I'm pretty sure the intermediate level monitoring should include normandy. is that for HPKP?

Yes, but it's not for HPKP (we don't look at HTTP headers), only for the ciphersuites configuration.

PR #229, opened 2017-07-24
Add fxpartners.mozilla.com to runner config

jasonthomas:
r?


jvehent:
Do you just want this one in the general section? Otherwise, I can add it to the 14 days paging and the TLS ciphersuites monitoring too.

jasonthomas:
I think having this in general section is sufficient as long as the cloud-ops team is notified (via email) that the certificate is going to expire. We didn't have coverage before and wanted to make sure it is monitored in the future.

PR #230, opened 2017-07-24
Add Dockerflow endpoints

autrilla:
Adds `/__heartbeat__`, `/__lbheartbeat__` and `/__version__` endpoints.

The `/__version__` endpoint will require some automation in CI to work, as the `version.json` file needs to be generated.

### DIFF
--- a/tlsobs-api/handlers.go
+++ b/tlsobs-api/handlers.go
@@ -8,6 +8,8 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"os"
+	"path"
 	"sort"
 	"strconv"
 	"time"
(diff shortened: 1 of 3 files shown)


tlsobs-api/handlers.go
```diff
@@ -636,6 +638,45 @@ func PreflightHandler(w http.ResponseWriter, r *http.Request) {
...
+	w.WriteHeader(http.StatusOK)
+}
+
+func versionHandler(w http.ResponseWriter, r *http.Request) {
+	dir, err := os.Getwd()
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	filename := path.Clean(dir + string(os.PathSeparator) + "version.json")
```
jvehent: Could you add that file to the repo to prevent the api from crashing when people run it directly?

autrilla: Done

tlsobs-api/handlers.go
```diff
@@ -636,6 +638,45 @@ func PreflightHandler(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func heartbeatHandler(w http.ResponseWriter, r *http.Request) {
+	val := r.Context().Value(ctxDBKey)
+	if val == nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	db, ok := val.(*pg.DB)
+	if !ok || db.Ping() != nil {
```
jvehent: `db.Ping()` doesn't actually talk to the database, so the preferred approach is to run a `SELECT 1`.
https://github.com/golang/go/issues/7619

tlsobs-api/handlers.go
```diff
@@ -636,6 +638,45 @@ func PreflightHandler(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
```
jvehent: can you add a small response body too? like `alive`. it's nicer for manual testing

tlsobs-api/handlers.go
autrilla: Yup, done. Looks like `Ping` only checks if the connection is alive

tlsobs-api/handlers.go
autrilla: Done

PR #234, opened 2017-07-28
Vendor newer scribe package and update references to it

ameihm0912:
- Vendor the newest scribe package
- Update references for EVR operations as the constant name has changed


ameihm0912:
@jvehent, these changes permit the newer package.

PR #235, opened 2017-07-28
Change dockerfile so that it follows dockerflow spec

autrilla:


Dockerfile
```diff
@@ -12,3 +12,12 @@ RUN go get github.com/mozilla/tls-observatory/tlsobs && \
     mv ./ev-checker /go/bin/ && \
     cd .. && \
     rm -rf ev-checker
+RUN addgroup -gid 10001 app && \
+    adduser --home /app --gecos "" --ingroup=app --uid=10001 --disabled-login app
+RUN mv $GOPATH/bin/tlsobs-api /app/ && \
+    mv $GOPATH/bin/tlsobs-scanner /app/ && \
+    mv $GOPATH/bin/tlsobs-runner /app/
```
jvehent: I think you need ev-checker in there as well

autrilla: Yup, added

PR #236, opened 2017-07-31
Make API listen address configurable

autrilla:
Not sure whether we want this to be configurable through the config file/environment variables or through command line arguments.


jvehent:
I'd prefer not adding another method of configuring the daemons. We already have a config file that's partially overridable via env variables. How about adding a new env variable to https://github.com/mozilla/tls-observatory/blob/master/config/config.go instead?

PR #243, opened 2017-08-08
Fix deploy_dockerhub

autrilla:
Fixes broken deploy_dockerhub script.

No need to check whether we're passing in `master`, as we always pass in `latest` and the commit hash.


autrilla:
I meant to just use `$1`. Other projects pass the current branch or tag to the script, but for tls-observatory I want to just use the commit hash as the docker image tag.

PR #244, opened 2017-08-08
Add scans and certificate metrics to API

autrilla:
We might also be interested in additional metrics such as scans processed, workers analysis executed, and number of trust relationships added, but that would have to run on the scanner, which doesn't have a way to pass arbitrary objects down the call stack like the `http` package allows with `context`.


tlsobs-api/handlers.go
```diff
@@ -17,8 +17,10 @@ import (
 	"bytes"
 	"crypto/sha256"
 
+	"github.com/aws/aws-sdk-go/service/cloudwatch"
```
jvehent: Why not put everything into the metrics package and not impact cloudwatch in the api/scanner?

tlsobs-api/main.go
```diff
@@ -71,17 +72,29 @@ func main() {
...
+		setResponseHeaders(),
+	}
+
+	sess, err := session.NewSession()
+	if err != nil {
+		log.Printf("Could not create AWS session: %s", err)
+	} else {
+		cloudwatchSvc := cloudwatch.New(sess)
+		middlewares = append(middlewares, addCloudwatch(cloudwatchSvc))
+		log.Printf("CloudWatch configured.")
```
jvehent: This might be cleaner in `middleware.go`.
Also, we should pref it to `off` for people running outside of aws.

autrilla: If setting up cloudwatch fails for any reason, we show a warning and keep going. I don't think we need a setting to turn it off if we do this.

autrilla:
I've moved the metric sending primarily to the database package, except where I can't because the scanner performs db queries itself. We should probably make all database interactions go through the database package, but I think it's best to leave that for another PR.

PR #245, opened 2017-08-10
Fix leading serial number zeroes not being included in the database

### LINKED ISSUE #218: Missing leading zero in serial number
We parse serial numbers of certificate into hex string stored in database. The parsing currently ignores leading zero, and should append them properly.

According to @jcjones, there should only ever be a single leading zero, as while ASN.1 permits INTEGER to do insane things, the WebPKI (because of DER-form) only lets there be one leading zero, corresponding to the first nybble of the most significant byte. In this case, the actual ASN.1 is:

02 04 **07 27 7F 52**

so it's cool and correct to show that leading zero. It's ASN.1-legal to encode something more like:

02 04 **00** 07 DE AD

but that's not the minimal form of the integer, so it is not DER-encoded, and thus shouldn't pass one of CertLint / CABLint (I forget which one).

@mozkeeler added:

Another important example here would be something like 02 03 **00 F0 0F** where the leading 00 indicates that this is a positive value, not negative. (Serial numbers must be positive under RFC 5280, but some CAs have issued certificates where the serial number's leading bit is set, meaning it is actually negative.)

A good rule might be to show exactly the contents of the value portion of the tag-length-value of the serial number encoding. That is, for the certificate in question, we might display "07277F52". For the example I
added, we might display "00F00F".

autrilla:
Fixes #218 

This patch changes the way we store the serial number in the database. Currently, we store the hex representation of the serial number. This patch stores the hex representation of the DER-encoded serial number. A few examples of what currently happens and what will happen after the patch:

| Serial Number (Decimal) | Currently stored | Stored after patch |
|-------------------------|------------------|--------------------|
| 1                       | 1                | 01                 |
| -1                      | -1               | FF                 |
| 0                       | 0                | 00                 |
| 201                     | C9               | 00C9               |
| -201                    | -C9              | FF37               |

You can try more examples here: https://play.golang.org/p/9BErIzLKCF

A possible improvement would be batching the updates in `tools/fixserialnumbers.go` to reduce the number of round trips.

### DIFF
--- a/certificate/certificate.go
+++ b/certificate/certificate.go
@@ -8,6 +8,7 @@ import (
 	"crypto/sha1"
 	"crypto/sha256"
 	"crypto/x509"
+	"encoding/asn1"
 	"encoding/base64"
 	"fmt"
 	"log"
@@ -458,7 +459,21 @@ func CertToStored(cert *x509.Certificate, parentSignature, domain, ip string, TS
 	stored.IPs = make([]string, 0)
 
 	stored.Version = cert.Version
-	stored.Serial = fmt.Sprintf("%X", cert.SerialNumber)
+
+	// Compute the ASN1 representation of the serial number
+	m, err := asn1.Marshal(cert.SerialNumber)
+	if err != nil {
+		stored.Serial = ""
+	}
+	// Unmarshal the ASN1 representation into a RawValue so we can trim
+	// the ASN1 tag and length
+	var rawValue asn1.RawValue
+	_, err = asn1.Unmarshal(m, &rawValue)
+	if err != nil {
+		stored.Serial = ""
+	}
+	stored.Serial = fmt.Sprintf("%X", rawValue.Bytes)
+
 	stored.SignatureAlgorithm = SignatureAlgorithm[cert.SignatureAlgorithm]
 
 	stored.Key, err = getPublicKeyInfo(cert)
--- a/certificate/certificate.go
+++ b/certificate/certificate.go
@@ -448,6 +448,20 @@ func getPublicKeyInfo(cert *x509.Certificate) (SubjectPublicKeyInfo, error) {
 
 }
 
+func GetHexASN1Serial(cert *x509.Certificate) (serial string, err error) {
+	m, err := asn1.Marshal(cert.SerialNumber)
+	if err != nil {
+		return
+	}
+	var rawValue asn1.RawValue
+	_, err = asn1.Unmarshal(m, &rawValue)
+	if err != nil {
+		return
+	}
+	serial = fmt.Sprintf("%X", rawValue.Bytes)
+	return
+}
+
 //certtoStored returns a Certificate struct created from a X509.Certificate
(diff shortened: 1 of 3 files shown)


jvehent:
This is a nice shortcut if doing a roundtrip through ASN1 does what we want. Would it be worthwhile to put the marshaling into a separate function and add tests to make sure we always get the expected values?

r? @jcjones & @mozkeeler

autrilla:
I think tests for this would be useful, yeah. I'll work on that.

tools/fixserialnumber.go
```diff
@@ -0,0 +1,188 @@
...
+				err: fmt.Errorf("Error updating serial number in database: %s", err),
+			}
+			continue
+		}
+		results <- result{id: j.id, err: nil, changed: true}
+	}
+}
+
+// getSerialNumberASN1Value returns the ASN1-encoded serial number for the given certificate, excluding the tag and length.
+func getSerialNumberASN1Value(cert *x509.Certificate) ([]byte, error) {
```
jvehent: I think this is the function we want to move to the certificate package. I'd make it `GetHexSerial` and return the hexadecimal string, since this is what we always manipulate.

autrilla: Yep, done. Less duplication this way.

certificate/certificate.go
```diff
@@ -447,6 +448,20 @@ func getPublicKeyInfo(cert *x509.Certificate) (SubjectPublicKeyInfo, error) {
 
 }
 
+func GetHexASN1Serial(cert *x509.Certificate) (serial string, err error) {
+	m, err := asn1.Marshal(cert.SerialNumber)
```
mozkeeler: Hmmm - presumably go's asn1 parser wouldn't parse the certificate if the serial number were encoded incorrectly in the first place? (e.g. if it had too much leading padding?) That's the only situation I can think of where round-tripping this through asn1 wouldn't result in the exact same bytes as encoded in the serial number of the certificate.

certificate/certificate.go
```diff
@@ -447,6 +448,20 @@ func getPublicKeyInfo(cert *x509.Certificate) (SubjectPublicKeyInfo, error) {
...
+	if err != nil {
+		return
+	}
+	var rawValue asn1.RawValue
+	_, err = asn1.Unmarshal(m, &rawValue)
+	if err != nil {
+		return
+	}
+	serial = fmt.Sprintf("%X", rawValue.Bytes)
```
mozkeeler: We've confirmed this 0-pads odd-length hex values? (e.g. if rawValue.Bytes is [0c, ff] this will result in "0CFF" and not "CFF"?)

autrilla: Yeah, see the tests: https://github.com/mozilla/tls-observatory/pull/245/files#diff-920a12e1044a9ee517527da2fff57753R21

certificate/certificate.go
autrilla: We parse certificates with `crypto/x509`'s `ParseCertificate` which blows up on a cert with incorrect encoding. Example by @jcjones here https://play.golang.org/p/shRibwS6_4

PR #246, opened 2017-08-16
Use a materialized view to retrieve statistics

jvehent:


april:
I do the same thing on the HTTP Observatory:

https://github.com/mozilla/http-observatory/blob/master/httpobs/database/schema.sql#L82

When you have 50B+ rows, there's just no way of doing these kinds of queries super quickly, indexes or not.  The only thing I might change is that I refresh the materialized view every so often, currently set to roughly every 10 minutes. It prevents returning stale data.

Not sure if this will kick off multiple REFRESHes if one is already running, but I'm pretty sure it obtains a lock and so it's not really a big deal.

Otherwise looks fine to me, aside from my Go-fu being relatively weak.

PR #251, opened 2017-08-21
Add Kubernetes deployment

autrilla:
This adds a simple Kubernetes configuration that runs postgres, the api and the scanner I used a few days ago to try to reproduce a bug.

 I don't know if we prefer docker-compose (#171) over kubernetes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN go install github.com/mozilla/tls-observatory/tlsobs-api && \
     go install github.com/mozilla/tls-observatory/tlsobs-scanner && \
     go install github.com/mozilla/tls-observatory/tlsobs-runner && \
     apt-get update -y && \
-    apt-get install git libcurl4-nss-dev libnss3 libnss3-dev clang -y && \
+    apt-get install git libcurl4-nss-dev libnss3 libnss3-dev clang postgresql-client -y && \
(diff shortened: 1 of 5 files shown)


jvehent:
That's cool, I guess. Though I'd prefer to put that in the doc than on its own set of files that will likely become out of date in a few months, since we don't actively use it.

Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN go install github.com/mozilla/tls-observatory/tlsobs-api && \
     go install github.com/mozilla/tls-observatory/tlsobs-scanner && \
     go install github.com/mozilla/tls-observatory/tlsobs-runner && \
     apt-get update -y && \
-    apt-get install git libcurl4-nss-dev libnss3 libnss3-dev clang -y && \
+    apt-get install git libcurl4-nss-dev libnss3 libnss3-dev clang postgresql-client -y && \
```
jvehent: Why?

autrilla: Needed a way to automate creation of the database and schema from the container. fwiw this doesn't increase the image size substantially. `docker images` shows a size of 1.34GB before this, and the same size after the change.

autrilla:
I added a warning on the README.md in kubernetes/ for what you mentioned.

PR #273, opened 2017-11-16
Updates for Firefox Accounts & Tiles

jbuck:


conf/runner.yaml
```diff
@@ -64,7 +64,7 @@ runs:
         - oauth.accounts.firefox.com
         - profile.accounts.firefox.com
         - verifier.accounts.firefox.com
-        - accounts.cdn.mozilla.net
```
jvehent: This still resolves to cloudfront. Is it no longer in use, and if so, can we get rid of the record in route53?

jbuck: Yep! I just deleted the CDN, S3 bucket, and route 53 record

floatingatoll:
Note that it's not fully deleted yet, in case that's a concern:

accounts.cdn.mozilla.net. 35    IN    CNAME    d2azfwpwj8hugi.cloudfront.net
.
mozilla.net.        62732    IN    NS    ns5-65.akam.net.

jbuck:
Haaaaah it's totally in Mozilla DNS too. Well that was a misleading DNS entry I deleted in Route 53!

jbuck:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1418446 to delete it within Mozilla's DNS. Thanks for the catch!

PR #276, opened 2017-11-17
Add conduit sites to runner

jvehent:


jvehent:
@ckolos: r?

ckolos:
We don't control bugzilla yet, so I would leave them off or commented out until we do. Otherwise, r+

Issue #226, opened 2017-07-08
Grade F???


v998:
Just ran the TLS Observatory test, however [it reports my website as insecure](https://observatory.mozilla.org/analyze.html?host=iniiter.com#tls), which I believe it should be a bug.

On my server I am using OpenSSL 1.0.2l together with [the Cloudflare Chacha20 patch](https://github.com/cloudflare/sslconfig/blob/master/patches/openssl__chacha20_poly1305_draft_and_rfc_ossl102j.patch)

And using the patched-OpenSSL mentioned above, I have the `ECDHE-RSA-CHACHA20-POLY1305`, `DHE-RSA-CHACHA20-POLY1305`, `ECDHE-RSA-CHACHA20-POLY1305-old` and `DHE-RSA-CHACHA20-POLY1305-old` ciphers configured on my httpd.

But Mozilla TLS Observatory is reporting these improvements for me:
> Consider adding ciphers ECDHE-ECDSA-CHACHA20-POLY1305, ECDHE-RSA-CHACHA20-POLY1305, ECDHE-ECDSA-AES128-GCM-SHA256, ECDHE-ECDSA-AES256-GCM-SHA384, ECDHE-ECDSA-AES128-SHA256, ECDHE-ECDSA-AES128-SHA, ECDHE-ECDSA-AES256-SHA384, ECDHE-ECDSA-AES256-SHA, ECDHE-ECDSA-DES-CBC3-SHA
Enforce server side ordering
Remove ciphersuites ECDHE-RSA-CHACHA20-POLY1305-old, DHE-RSA-CHACHA20-POLY1305-old

So here comes the issue:
1. `ECDHE-RSA-CHACHA20-POLY1305-old` or `ECDHE-RSA-CHACHA20-POLY1305-old` are NOT insecure, they are only the draft versions and are currently used by Cloudflare. But TLS Observatory is telling me to remove these ciphers (so TLS Observatory seems to be thinking these ciphers are insecure..)

[...]

Issue #186, opened 2016-12-19
Cloudflare sites all return as "Insecure"


marumari:
`threatresponse.cloud` is a site that has this issue, I think it's because of having the cipher ECDHE-ECDSA-CHACHA20-POLY1305-OLD, but I'm not entirely positive.

marumari:
Also I believe it supports both CHACHA20-OLD and CHACHA20, but the TLS Observatory only shows the former.

jvehent:
We'll need to add `ECDHE-ECDSA-CHACHA20-POLY1305-OLD` to the JSON recommendations to fix that. Support for scanning CHACHA20 is being discussed here: https://github.com/mozilla/cipherscan/issues/118

grittygrease:
Can the policy be updated to just ignore support for ECDHE-ECDSA-CHACHA20-POLY1305-OLD in the rating determination? It's widely assumed that there are no significant security differences between this cipher and the RFC version. We support both. A large percentage of ChaCha20-enabled clients use this cipher (mostly older Chrome versions), so we are not inclined to disable support for it without a strong reason.

This issue is currently affecting millions of sites who are running on Cloudflare using our configuration, which we consider to be among the most secure configurations possible with broad compatibility.

marumari:
So the trick here is that to make it part of "Modern", we would need to make it part of our Modern configurations:

https://wiki.mozilla.org/Security/Server_Side_TLS
https://github.com/mozilla/server-side-tls/blob/gh-pages/json/server-side-tls-conf.json

And I don't know if we want to recommend that people put the old code point in their server configurations.  Could we make a one-off exception in cipherscan/tls-observatory to treat it as the same as ChaCha20, instead of modifying the upstream recommendations?

jvehent:
Blocked on https://github.com/mozilla/server-side-tls/issues/178
I'll try to get to it next week.

jvehent:
I put a wart in the code to deal with this: https://github.com/mozilla/tls-observatory/commit/390a0500e0c3107845d08fe2747a702b1a461044
Now the config returns "non compliant" because it is somewhere in between modern and intermediate.

To get to modern, it would need:
  - for modern level: remove ciphersuites ECDHE-ECDSA-AES128-SHA, ECDHE-ECDSA-AES256-SHA, ECDHE-ECDSA-DES-CBC3-SHA
[...]

PR #280, opened 2017-12-05
Add exception for RSA CHACHA20 cipher to fix cloudflare sites with RSA certs being marked as insecure

johnguant:
Add the ECDHE-RSA-CHACHA20-POLY1305-OLD cipher to the ignored ciphers added for #186 as this is also used by cloudflare on sites that have a RSA certificate as noted in #226


jvehent:
Yep, I had missed that one... thanks for the patch, it'll make @grittygrease happy :)

PR #286, opened 2018-01-05
Added OCSP worker to retrieve certificate status

fr0zenbits:
This is one of my first go projects, so any constructive feedback is appreciated. I know the OCSP code works, but I am not sure if I implemented the worker-releated code correctly.


worker/ocspStatusWorker/ocspStatusWorker.go
```diff
@@ -0,0 +1,83 @@
...
+
+	res <- worker.Result{
+		Success:    true,
+		WorkerName: workerName,
+		Errors:     nil,
+		Result:     out,
+	}
+}
+
+func (w ocspStatusWorker) error(res chan worker.Result, messageFormat string, args ...interface{}) {
```
0xdiba: I don't think a separate error function is needed.
You could just initialize a new worker.Result in the beginning, in case of error just append to the Errors slice and in case of success just marshal the result in.

worker/ocspStatusWorker/ocspStatusWorker.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (w ocspStatusWorker) Run(in worker.Input, res chan worker.Result) {
+	out, err := json.Marshal(in.Params)
+
+	conn, err := tls.Dial("tcp", in.Connection.ScanIP + ":443", nil)
+	if err != nil {w.error(res, "Could not connect to server: %s", err)}
+	defer conn.Close()
+
+	issuerCertificate := conn.ConnectionState().PeerCertificates[1]
+	certificate := conn.ConnectionState().PeerCertificates[0]
```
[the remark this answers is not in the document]
fr0zenbits: So the `worker.Input.CertificateChain.Certs` is a string array. How do you covert the strings in the array to a x509 certificate? I need to pass in the website/issuer cert as a x509 when creating the ocsp request.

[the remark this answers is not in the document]
0xdiba: Each entry in the array is a base64 encoding of the raw certificate so all you have to do is :
```
raw := base64.DecodeString(cert)
cert := x509.ParseCertificate(raw)
```

fr0zenbits:
How's this?

worker/ocspStatusWorker/ocspStatusWorker.go
```diff
@@ -0,0 +1,81 @@
+package ocspStatusWorker
```
jvehent: nit: remove "Worker" from the name, it's already under /worker/

worker/ocspStatusWorker/ocspStatusWorker.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type params struct {
+}
+
+func (w ocspStatusWorker) Run(in worker.Input, resChan chan worker.Result) {
+	res := worker.Result{WorkerName: workerName, Success:false}
+
+	rawCert, _ := base64.StdEncoding.DecodeString(in.CertificateChain.Certs[0])
+	certificate, _ := x509.ParseCertificate(rawCert)
```
jvehent: the end entity is already parsed in `in.Certificate`

worker/ocspStatusWorker/ocspStatusWorker.go
```diff
@@ -0,0 +1,81 @@
...
+	rawIssuerCert, _ := base64.StdEncoding.DecodeString(in.CertificateChain.Certs[1])
+	issuerCertificate, _ := x509.ParseCertificate(rawIssuerCert)
```
jvehent: You should check for errors here.

It might also be worthwhile to make sure the issuer is the right one (that it sign the end-entity) because ordering isn't always guaranteed. You can use CheckSignatureFrom from the x509 package for this.

worker/ocspStatusWorker/ocspStatusWorker.go
```diff
@@ -0,0 +1,81 @@
...
+
+	opts := &ocsp.RequestOptions{Hash: crypto.SHA256}
+	req, err := ocsp.CreateRequest(certificate, issuerCertificate, opts)
+	if err != nil {
+		res.Errors = append(res.Errors, err.Error())
```
jvehent: errors need to be inserted into the response channel, otherwise they will be dropped. Something like this:
```go
func (w runner) error(res chan worker.Result, messageFormat string, args ...interface{}) {
        out, _ := json.Marshal(fmt.Sprintf(messageFormat, args...))
        res <- worker.Result{
                Success:    false,
                WorkerName: workerName,
                Result:     out,
        }
}
```
and use it as follows:
```go
        if err != nil {
                w.error(res, "there was an error: %v", err)
                return
        }
```

worker/ocspStatusWorker/ocspStatusWorker.go
```diff
@@ -0,0 +1,81 @@
...
+		return
+	}
+
+	httpResponse, err := http.Post(certificate.OCSPServer[0], "application/ocsp-request", bytes.NewReader(req))
+	if err != nil {
+		res.Errors = append(res.Errors, err.Error())
+		return
+	}
+
```
jvehent: add a `defer httpResponse.Body.Close()`

worker/ocspStatusWorker/ocspStatusWorker.go
```diff
@@ -0,0 +1,81 @@
...
+	OCSPResponse, err := ocsp.ParseResponse(output, issuerCertificate)
+	if err != nil {
+		res.Errors = append(res.Errors, err.Error())
+		return
+	}
+
+	status := OCSPStatus{ Status:OCSPResponse.Status, RevokedAt:OCSPResponse.RevokedAt }
+
+	out, _ := json.Marshal(status)
```
jvehent: don't ignore errors, you never know what the OCSP response will contain

fr0zenbits:
See if that's better

jvehent:
@fr0zenbits : you say it's ready to be reviewed again but I'm not seeing any changes following my review. Are there missing commits?

adamdecaf:
@fr0zenbits hey! I took your work in this PR and finished out the OCSP status worker. Hope that's ok. See https://github.com/mozilla/tls-observatory/pull/336

Issue #271, opened 2017-11-08
The TLS-Observatory fails to parse certificates that have PrintableString fields with invalid ASN.1 characters.


christopher-henderson:
The [ASN.1 Specification for PrintableString](https://www.itu.int/ITU-T/studygroups/com17/languages/X.680-0207.pdf#page=67&zoom=auto,-22,812) allows for only a small set of characters. As a result, the following two certificates from the CCADB cannot be parsed by the Observatory:

Error:
`Could not parse X.509 certificate: asn1: syntax error: PrintableString contains invalid character`

[...]

Issue #283, opened 2018-01-04
The CRL Distribution list of some intermediate certificates are being truncated


christopher-henderson:
The following intermediates pulled from the CCADB are presenting with the issue. I have tracked it down to an issue in the golang encoding/asn1 package. Go is looking at the second byte of each field for the content length of that field, however these certs appear to be encoding the length of the _entire list_ in that particulasr byte while encoding the length of the individual field in the final byte - causing the asn1 library to think that there is only one element. I'm still working through it, however, so this may be a red herring.

9858CEB7B72F296B9C8CEE355F92B4E0D18DECAAF2C3696E6244A2E203EF6345
&nbsp;&nbsp;[crt.sh](http://crt.sh/?q=9858CEB7B72F296B9C8CEE355F92B4E0D18DECAAF2C3696E6244A2E203EF6345)
&nbsp;&nbsp;[TLS Observatory](https://tls-observatory.services.mozilla.com/static/certsplainer.html/?sha256=9858CEB7B72F296B9C8CEE355F92B4E0D18DECAAF2C3696E6244A2E203EF6345)
B01A8DC9426D48767157FFCB46BEFA193175BBC981B57D47BC6C7FF71B94A35C
&nbsp;&nbsp;[crt.sh](http://crt.sh/?q=B01A8DC9426D48767157FFCB46BEFA193175BBC981B57D47BC6C7FF71B94A35C)
&nbsp;&nbsp;[TLS Observatory](https://tls-observatory.services.mozilla.com/static/certsplainer.html/?sha256=B01A8DC9426D48767157FFCB46BEFA193175BBC981B57D47BC6C7FF71B94A35C)
9858CEB7B72F296B9C8CEE355F92B4E0D18DECAAF2C3696E6244A2E203EF6345
&nbsp;&nbsp;[crt.sh](http://crt.sh/?q=9858CEB7B72F296B9C8CEE355F92B4E0D18DECAAF2C3696E6244A2E203EF6345)
&nbsp;&nbsp;[TLS Observatory](https://tls-observatory.services.mozilla.com/static/certsplainer.html/?sha256=9858CEB7B72F296B9C8CEE355F92B4E0D18DECAAF2C3696E6244A2E203EF6345)
592C508E39D7665F844586C3B9E765AFD5D80BBE05F9289CB0DED5711AC9C10A
[...]

Issue #281, opened 2017-12-06
Certs with `Name Constraints` critical constraint fails to parse


christopher-henderson:
The following certificates fail to parse due to:

`
x509: unhandled critical extension
`

https://crt.sh/?id=261854056&opt=x509lint
https://crt.sh/?id=244671923&opt=x509lint
https://crt.sh/?id=244671921&opt=x509lint
https://crt.sh/?id=12729480&opt=x509lint
https://crt.sh/?id=163322577&opt=x509lint
https://crt.sh/?id=215376214&opt=x509lint
https://crt.sh/?id=244671919&opt=x509lint
https://crt.sh/?id=163688490&opt=x509lint

This is being cause by the Name Constraint extension which the current version of go (1.9.2) is not handling. The `devel` branch of go, however, is handling this just fine. I'll update this issue with a patch number after figuring out the git blame on it so that we can figure out what version this will start working in.

christopher-henderson:
Adam Langely appears to have patched this in [9e76ce70701ceef8fbccfb953b33a2ae7fe0367c](https://github.com/golang/go/commit/9e76ce70701ceef8fbccfb953b33a2ae7fe0367c), which is marked as a 1.10 milestone. So we should hopefully see this is release in February 2018.

PR #288, opened 2018-01-11
Build the Observatory using Go 1.11

christopher-henderson:
This fixes issues #271, #281, and #283.


jvehent:
Ha, I had circleci configured to not build PRs from forks. Fixed now. Could you rebase and repush, which should trigger a build.

jvehent:
ok, rebased through github, waiting on CI now

jvehent:
Looks like circleci doesn't know Go 1.11 yet...
`Error response from daemon: manifest for circleci/golang:1.11 not found`

christopher-henderson:
Yeah, I was afraid of that. Nuts. I'll see about rolling an image onto Docker Hub.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:latest
+FROM nvor/golang:2dc025e4e1
```
jvehent: errrr... not sure if I'm OK with using a 3rd party image instead of an official one here. This is going to run in our production infra, after all. How about building go1.11 ourselves?

christopher-henderson: Sure, yeah, that's what I did. This image is mine (chenderson was taken so I took an old internet alias of mine - Nvor). But it's not a Mozilla account, so fair enough.

I'd suggest either pulling my image and recomitting to your Docker Hub, or you can produce your own doing what I did:

```bash
$ docker run --name golang golang /bin/bash -c "git clone https://github.com/golang/go.git /tmp/go; cd /tmp/go/src/; ./all.bash; rm -rf /usr/local/go; mv /tmp/go /usr/local; go version"

$ docker run --name circleci_golang circleci/golang /bin/bash -c "git clone https://github.com/golang/go.git /tmp/go; cd /tmp/go/src/; ./all.bash; sudo rm -rf /usr/local/go; sudo mv /tmp/go /usr/local; go version"
```

jvehent: I think we should use a basic Alpine image and install go ourselves using the commands above. Want to give it a shot? Otherwise I'll try to get to it.

jvehent:
@christopher-henderson Should we close this PR for now?

jcjones:
You'll need to make sure that this supersedes https://github.com/mozilla/tls-observatory/pull/333

adamdecaf:
Are there known changes already in Go 1.11 that we need here? I see a couple linked issues, but don't know the context if Go 1.10 solves any of them.

Dockerfile
christopher-henderson: SO, building Go on Alpine has had know difficulties since last fall ([build: get Alpine builders passing](https://github.com/golang/go/issues/19938)).  I'm specifically get crit in the face by this [SIGSEV in CGO](https://github.com/golang/go/issues/19938#issuecomment-297243685). Seems like they use a specifically patched version of Go to build on Alpine. For the sake of reproducibility, here's the Dockerfile I have so far.

```
FROM golang:alpine
RUN apk add --no-cache git bash gcc build-base
RUN git clone https://github.com/golang/go.git /tmp/go
RUN cd /tmp/go/src/; git checkout 4458a357ab819a612c0c4cafae88a65287254fe9; ./all.bash
RUN rm -rf /usr/local/go
RUN mv /tmp/go /usr/local
RUN rm -rf /tmp/*
RUN apk del git bash gcc build-base
```

I think I'm going to continue to try to get the previous image smaller. It's based off of docker.io/golang:latest. I'm not a Docker wizard, although it's frustrating that even after cleaning up build artifacts and squashing the filesystem it ends over twice as big as the base image.

jvehent: Compress all of those `RUN` into a single one joined with `&&` and it'll remove 6 layers from your image.

christopher-henderson: Thought I'd give that a whack, although it's getting the same result as before as I was using `--squash` anyways. Although still to good to know that.

christopher-henderson: Argh, I got alpine building by stealing the Go team's own Dockerfile which applies the necessary patches, but the Observatory is banking on a Debian-like environment for things like `apt-get`. Trying again...

christopher-henderson: Alright, I have the two images down to the following sizes:

Go: 850 MB, as compared to the official image's size of 780 MB
CircleCI: 1.17 GB, as a compared to the official image's size of 1.03 GB

Several interesting side notes here.

- When building these images I used git to pull the golang repo which is...a git repo. Git repos have .git directories. .git directories are big. Delete those .git directories. Duh.
- `FROM` lines do not have their layers squashed. The biggest gains I got in terms of size was to expand `FROM` directives into copy-pasting the entire original Dockerfile, coupled with the `--squash` flag.
- Use `make.bash` instead of `all.bash`. `all.bash` is nice since it runs all of the unit tests, but it also builds and packages hundreds of MB of unnecessary stuff. `make.bash` is what the Go team uses to deploy.

These gists are the dockerfiles I used to create these images, in case ya'll would rather build and vendor these under a Mozilla name.

- [Go 1.11](https://gist.github.com/christopher-henderson/8af00c106703da0439b3bde7b978dd86)
- [CircleCI, Go 1.11](https://gist.github.com/christopher-henderson/ca078a75403a5ead1ffe6eea3768e4d3)

jvehent:
I'm going to close this in favor of #343 which builds go1.11 on top of the standard golang:latest container.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:latest
+FROM nvor/golang:4458a35
```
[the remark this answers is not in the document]
christopher-henderson: Yup, that's fine. It's exactly why I linked you the gists of how these were made. =)

PR #295, opened 2018-02-04
Support scanning non-standard ports

### LINKED ISSUE #292: Add a configuration flag to allow scanning non-standard ports
I work for a company looking to deploy the Observatory project to test it's internal servers. Some of these servers run on non-standard ports.

My intention is that this is an optional, disabled-by-default feature which would allow internal deployments to scan non-standard ports; but for most deployments to not have to worry about it being an issue.

software-opal:
Fixes #292

Currently makes a minimal change to support scanning non-standard ports. Currently the results appear to be as-expected.

I'm experiencing problems with `ev-checker`. I think it expects a domain name without a port number and I'm giving it one with a port number.

I'd also appreciate some guidance on how to implement the configuration option so that scanning can be restricted to a subset of ports. My understanding is that currently there is no support for filtering the scanning.


autrilla:
As I understand it `ev-checker` checks that the CN of the retrieved certificate matches whatever is passed as the hostname, so if it's passed with the port, the check will fail.

I'd implement the filtering of scanned ports with a comma separated list of allowed ports in the config (https://github.com/mozilla/tls-observatory/blob/master/conf/scanner.cfg, parsed here https://github.com/mozilla/tls-observatory/blob/master/config/config.go#L35). Note that the configuration loading code is shared by all components of the TLS Observatory.

jvehent:
> As I understand it ev-checker checks that the CN of the retrieved certificate matches whatever is passed as the hostname, so if it's passed with the port, the check will fail.

It's probably best to skip the `ev-checker` if the port is not 443.

jvehent:
> I'm experiencing problems with ev-checker. I think it expects a domain name without a port number and I'm giving it one with a port number.

This should now be fixed in master: ev-checker won't run if a port is specified.

jvehent:
@leesdolphin Could you rebase this patch with master?

Issue #298, opened 2018-02-23
Warn on a certificate that chain to Symantec roots


jcjones:
Trust paths which anchor to [Symantec roots in this list](https://searchfox.org/mozilla-central/source/security/certverifier/TrustOverride-SymantecData.inc) are generally distrusted in Firefox 60 (based on a `notBefore` check) or Firefox 63 (no `notBefore` check). There's also a whitelist [of some intermediates' SPKIs from Apple, Google, and DigiCert](https://searchfox.org/mozilla-central/source/security/certverifier/TrustOverride-AppleGoogleDigiCertData.inc).

It would be nice to provide a warning and a link to [the Upcoming_Distrust_Actions wiki](https://wiki.mozilla.org/CA/Upcoming_Distrust_Actions) either:
1) If there's a Symantec root at the anchor, and a message saying "this might not work if it's not from one of the narrow whitelists", or
2) Apply the Symantec blacklist and the SPKI whitelist and be very clear that this is broken or about to be broken.

The `notBefore` check seems unnecessary to duplicate, as all of these certs need to change.

jcjones:
Another way to handle this is by blacklist of the actual end entity certificates. I believe Sleevi is producing one, so could just do a lookup of those, as we have a complete list.

jvehent:
If a cert has two paths, and only one goes through a distrusted root, what should the result of the analysis be?

floatingatoll:
A browser would, assuming removal of the Symantec intermediary, treat that
as a trusted certificate if the second path is otherwise trusted.

Are unexpired certificates with two paths, one distrusted, prevalent in the
wild?

jvehent:
for archiving, the sql queries that produce the report:
```sql
SELECT COUNT(DISTINCT(target)) AS "Sample size"
FROM scans
  INNER JOIN analysis ON (scans.id=analysis.scan_id)
  INNER JOIN certificates ON (scans.cert_id=certificates.id)
WHERE has_tls=true
  AND worker_name='symantecDistrust'
  AND timestamp > '2018-05-16 08:00:00';

SELECT COUNT(DISTINCT(target)) AS "Distrusted in 60"
FROM scans
  INNER JOIN analysis ON (scans.id=analysis.scan_id)
  INNER JOIN certificates ON (scans.cert_id=certificates.id)
WHERE has_tls=true
  AND worker_name='symantecDistrust'
  AND timestamp > '2018-05-16 08:00:00'
[...]

PR #299, opened 2018-02-24
Add symantec distrust evaluation worker

jvehent:
This mostly works, with a few quirks I'm still ironing out.

It probably needs to be smarter about evaluating the paths, and making sure that a valid path is only valid if it maps to a trusted root. Which brings the question of which root do we trust? (probably mozilla's)

For now, testing `tesla.com` returns the following:
> "not impacted, but blacklisted certs found in path: /C=US/O=VeriSign, Inc./OU=VeriSign Trust Network/OU=(c) 2006 VeriSign, Inc. - For authorized use only/CN=VeriSign Class 3 Public Primary Certification Authority - G5"

### DIFF
--- a/worker/symantecDistrust/symantecDistrust.go
+++ b/worker/symantecDistrust/symantecDistrust.go
@@ -1,0 +1,175 @@
+package symantecDistrust
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/mozilla/tls-observatory/certificate"
+	"github.com/mozilla/tls-observatory/logger"
+	"github.com/mozilla/tls-observatory/worker"
+)
+
+var (
+	workerName = "symantecDistrust"
+	workerDesc = "Checks if the target is impacted by https://wiki.mozilla.org/CA/Upcoming_Distrust_Actions"
+	log        = logger.GetLogger()
+)
+
+func init() {
+	runner := new(runner)
+	worker.RegisterWorker(workerName, worker.Info{Runner: runner, Description: workerDesc})
+}
+
+type runner struct {
+}
+
+var blacklist = []distrustedCert{
+	{
+		Name:   "/C=US/O=GeoTrust Inc./CN=GeoTrust Global CA",
+		SHA256: "FF856A2D251DCD88D36656F450126798CFABAADE40799C722DE4D2B5DB36A73A",
+	},
+	{
+		Name:   "/C=US/O=GeoTrust Inc./OU=(c) 2007 GeoTrust Inc. - For authorized use only/CN=GeoTrust Primary Certification Authority - G2",
+		SHA256: "5EDB7AC43B82A06A8761E8D7BE4979EBF2611F7DD79BF91C1C6B566A219ED766",
+	},
+	{
+		Name:   "/C=US/O=GeoTrust Inc./OU=(c) 2008 GeoTrust Inc. - For authorized use only/CN=GeoTrust Primary Certification Authority - G3",
+		SHA256: "B478B812250DF878635C2AA7EC7D155EAA625EE82916E2CD294361886CD1FBD4",
+	},
+	{
+		Name:   "/C=US/O=GeoTrust Inc./CN=GeoTrust Primary Certification Authority",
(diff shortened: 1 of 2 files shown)


worker/symantecDistrust/symantecDistrust.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func evalPaths(paths certificate.Paths) (trust bool, reason string) {
+	blacklisted, name := evalBlacklist(paths.Cert.Hashes.SHA256)
+	if blacklisted {
+		return false, name
+	}
+	for _, parent := range paths.Parents {
+		var theirReason string
+		trust, theirReason = evalPaths(parent)
```
jcjones: The nature of this is that `trust` will be set `true` or `false` depending on the final parent checked, which is not likely what we want. It seems like we want it gated: either it's `true` if `evalPaths()` ever returns true, or it's the pessimistic case of  `false` if `evalPaths()` ever returns false. I'm not sure the UI implications of these choices, but IMO it should latch.

jvehent:
How much detail do you want in the result? I could return the list of paths
that successes and failed, and why, just a Boolean, and anything in between.

jcjones:
I think it'd be fine to just be boolean, where if any of the paths are affected, we give a link to the wiki page?

jvehent:
Well, I was curious about the details, so I added some :) New output is
```
* Symantec distrust: impacted
  - path uses a blacklisted cert: C=US, O=VeriSign, Inc., OU=VeriSign Trust Network, OU=(c) 2006 VeriSign, Inc. - For authorized use only, CN=VeriSign Class 3 Public Primary Certification Authority - G5 (id=147)
  - path uses a root not trusted by Mozilla: C=US, O=VeriSign, Inc., OU=Class 3 Public Primary Certification Authority (id=154)
  - path uses a root not trusted by Mozilla: C=US, O=VeriSign, Inc., OU=Class 3 Public Primary Certification Authority (id=157)
  - for details, read https://wiki.mozilla.org/CA/Upcoming_Distrust_Actions
```
I inversed some of the logic so trust became distrust, and added a truststore check to only return impacted when a path of a root trusted by Mozilla cannot be found. That should be close enough to what Firefox does.

Remains the question of the whitelist you mentioned in #298. What, if anything, should we do about this?

jvehent:
Google.com shows as impacted, probably because we don't apply the whitelist?
```
* Symantec distrust: impacted
  - path uses a root not trusted by Mozilla: C=US, O=Equifax, OU=Equifax Secure Certificate Authority (id=63)
  - path uses a blacklisted cert: C=US, O=GeoTrust Inc., CN=GeoTrust Global CA (id=66)
  - for details, read https://wiki.mozilla.org/CA/Upcoming_Distrust_Actions
```

jvehent:
Id like to implement exactly the same logic as Firefox, so the whitelist is
important.

How does it work? I imagine it overwrites the distrust decision, but how?

jvehent:
@jcjones does this look correct for google.com ?

```
--- Trust paths ---
C=US, O=Google Inc, CN=*.google.com (id=395)
└──C=US, O=Google Inc, CN=Google Internet Authority G2 (id=396)
   ├──C=US, O=GeoTrust Inc., CN=GeoTrust Global CA (id=397)
   │  └──C=US, O=Equifax, OU=Equifax Secure Certificate Authority (id=63)
   └──C=US, O=GeoTrust Inc., CN=GeoTrust Global CA (id=66)

* Symantec distrust: not impacted, but found paths with distrusted certs
  - path uses a blacklisted cert: C=US, O=GeoTrust Inc., CN=GeoTrust Global CA (id=66)
  - whitelisted intermediate C=US, O=Google Inc, CN=Google Internet Authority G2 (id=396) override blacklisting of 66
  - path uses a root not trusted by Mozilla: C=US, O=Equifax, OU=Equifax Secure Certificate Authority (id=63)
  - whitelisted intermediate C=US, O=Google Inc, CN=Google Internet Authority G2 (id=396) override blacklisting of 397
  - for details, read https://wiki.mozilla.org/CA/Upcoming_Distrust_Actions
```

PR #302, opened 2018-03-02
Timeout cipherscan after 3 minutes

jvehent:
Trying to prevent scans from hanging forever, which slows down scanners. I also added some code to expose the timeout information in the cmdline tool through the API.

### DIFF
--- a/tlsobs-scanner/main.go
+++ b/tlsobs-scanner/main.go
@@ -205,11 +205,13 @@ func (s scanner) scan(scanID int64, cipherscan string) {
 	// Cipherscan the target
 	js, err := connection.Connect(scan.Target, cipherscan)
 	if err != nil {
-		err, ok := err.(connection.NoTLSConnErr)
+		_, ok := err.(connection.NoTLSConnErr)
 		if ok {
 			//does not implement TLS
 			db.Exec("UPDATE scans SET has_tls=FALSE, completion_perc=100 WHERE id=$1", scanID)
 		} else {
+			//does not implement TLS
+			db.Exec("UPDATE scans SET scan_error=$1, completion_perc=100 WHERE id=$2", err.Error(), scanID)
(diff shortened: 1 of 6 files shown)


tlsobs-scanner/main.go
```diff
@@ -205,11 +205,13 @@ func (s scanner) scan(scanID int64, cipherscan string) {
 	// Cipherscan the target
 	js, err := connection.Connect(scan.Target, cipherscan)
 	if err != nil {
-		err, ok := err.(connection.NoTLSConnErr)
+		_, ok := err.(connection.NoTLSConnErr)
 		if ok {
 			//does not implement TLS
 			db.Exec("UPDATE scans SET has_tls=FALSE, completion_perc=100 WHERE id=$1", scanID)
 		} else {
+			//does not implement TLS
```
0xdiba: Invalid comment.

tlsobs/main.go
```diff
@@ -118,6 +118,10 @@ getresults:
 		if err != nil {
 			panic(err)
 		}
+		if results.Complperc == 100 && results.ScanError != "" {
+			fmt.Printf("Scan failed with error: %s\n", results.ScanError)
+			os.Exit(81)
```
0xdiba: Should this be documented in the README?
Since we are differentiating exit codes ( so not considering this a web page fault ) we could document it in case anybody is using tlsobs as part of CI/Sec scan.

tlsobs-scanner/main.go
jvehent: Good catch!

tlsobs/main.go
jvehent: Probably, but honestly I'd only bother if someone asks for it. Otherwise, it's easy enough to read the source.

0xdiba: Sounds reasonable.
Everything looks good to merge!

PR #306, opened 2018-03-08
Rename package use of Sirupsen to sirupsen

### LINKED ISSUE #305: Vendored version of logrus uses incorrect import path
The current way in which Logrus is vendored in this project breaks the new golang.org dep tool (see for more details: https://github.com/golang/dep/issues/1738). 

The issue: the vendored version is `github.com/Sirupsen/logrus` and it should be `github.com/sirupsen/logrus`.

From the Logrus README: 
> Seeing weird case-sensitive problems? It's in the past been possible to import Logrus as both upper- and lower-case. Due to the Go package environment, this caused issues in the community and we needed a standard. Some environments experienced problems with the upper-case variant, so the lower-case was decided. Everything using logrus will need to use the lower-case: github.com/sirupsen/logrus. Any package that isn't, should be changed.

For more details: 
https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276

Is it possible to correct your vendored dependencies?

gcmurphy:
Fixes #305

Also remove the Sirupsen vendored version (included twice).


gcmurphy:
Closing as fixed by maintainer.

jvehent:
Ha, sorry for the collision.

Issue #212, opened 2017-02-22
Use CA CRLs to check revocation status


jvehent:
Loads of certificates are revoked before the end of their validity period. We should check for revocation status using the CRLs provided by CAs, and maybe even run OCSP queries.

The CRLs should be reloaded regularly. We can either keep them in memory or load them in a new table.

Probably the best way to implement this is in a separate worker that modifies the certificate and trust tables.

0xdiba:
Fixed with 8791a200eb40f8625a152bfb8336171305f5f35c

PR #309, opened 2018-03-10
crlWorker: initial setup to retrieve CRL data

adamdecaf:
I wanted to test out what adding a worker would be like. This is a quick example for pulling CRL data.

Couple questions: 

1. Is this the right pattern so far? 
1. Caching / batch jobs - Would the idea be to store CRL responses in postgres?
   1. New table with `crl_distribution_point`, `fetched_at` and `response` ?
   1. Do we care about errors? CRL points have uptime objectives which we could track.

Issue: https://github.com/mozilla/tls-observatory/issues/212


worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,117 @@
...
+	var wg sync.WaitGroup
+	var out []byte
+
+	crlPoints := cert.X509v3Extensions.CRLDistributionPoints
+	wg.Add(len(crlPoints))
+	for i := range crlPoints {
+		go func(point string, wg *sync.WaitGroup) {
+			defer wg.Done()
+
+			resp, err := http.Get(point)
```
jvehent: I'd prefer to have an HTTP Transport with timeouts, probably around 30s.

worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,117 @@
...
+				return
+			}
+			defer resp.Body.Close()
+			bs, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				return
+			}
+			// TODO(adam): potential race here, but do we care?
+			// Would different CRL responses return different results?
+			out = bs
```
jvehent: why not store all CRL results instead?

worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return nil, fmt.Errorf("Unable to load CRL data for certificate %s", cert.Subject)
+}
+
+// Assertor compares two crlWorker Results and reports differences
+func (eval) Assertor(crlResult, assertresults []byte) (pass bool, body []byte, err error) {
+	// json.Unmarshal
+
+	return false, nil, nil // TOOD(adam)
+}
```
jvehent: it's best to not implement the interface at all than to leave an empty stub

worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,117 @@
...
+func (e eval) Run(in worker.Input, resChan chan worker.Result) {
+	result := worker.Result{
+		WorkerName: workerName,
+	}
+	crlRes := Result{}
+
+	// Grab first CRL response we can from our certificate
+	crlResp, err := e.getFirstCRLData(in.Certificate)
+	if err != nil {
+		result.Errors = append(result.Errors, err.Error())
```
jvehent: all of those errors should force the worker to exit immediately, something along the lines of 
```go
        if err != nil {
                res <- worker.Result{
                        Success:    false,
                        WorkerName: workerName,
                        Errors:     []string{fmt.Sprintf("failed to foo bar: %v", err)},
                        Result:     nil,
                }
        }
```

worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// Verify our CRL response is signed properly
+	x509Cert, err := in.Certificate.ToX509()
+	if err != nil {
+		result.Errors = append(result.Errors, err.Error())
+	}
+	if err := x509Cert.CheckCRLSignature(certList); err != nil {
```
jvehent: That's the wrong logic, I think. The CRL will be signed by the root or intermediate that issued it, not by the end-entity that is being revoked. The CA doesn't even have access to the private key of the end-entity. If you want to verify the CRL signature, you need to walk the path up from the end-entity to the root(s).

worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,117 @@
...
+		result.Errors = append(result.Errors, err.Error())
+	}
+
+	// Check if our certificate is in the revoked list
+	revoked := certList.TBSCertList.RevokedCertificates
+	for i := range revoked {
+		if x509Cert.SerialNumber.Cmp(revoked[i].SerialNumber) == 0 {
+			// certificate is in revoked list, serials match
+			crlRes.Revoked = true
```
jvehent: it's be interesting to also store the revocation time

worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,117 @@
...
+	// Marshal the response
+	bs, err := json.Marshal(&crlRes)
+	if err != nil {
+		result.Success = false
+		result.Errors = append(result.Errors, err.Error())
+	} else {
+		result.Success = true
+		result.Result = bs
+	}
+	resChan <- result
```
jvehent: I think it'd be useful to add a column to the certificate table to store the fact it has been revoked via CRL.

jvehent:
On caching: instead of doing it in postgres, I'd propose to do it in the scanner itself, with an in-memory cache. We already do this for certificate paths [1](https://github.com/mozilla/tls-observatory/blob/master/database/database.go#L19) [2](https://github.com/mozilla/tls-observatory/blob/master/database/certificate.go#L609-L616).

On errors: the analysis results should contain the error, and the printer should display them, and we can run queries on that if needed.

adamdecaf:
> I think it'd be useful to add a column to the certificate table to store the fact it has been revoked via CRL. 

> why not store all CRL results instead?

Should I just create the `alter table add column` and new table to `database/schema.sql`? I don't see any usage of a migration tool.

jvehent:
We don't have a migration tool, though I've been thinking about adding one for too long now.
I generally just add the column to schema.sql and run the alter table manually in our hosted environments.

adamdecaf:
Do you think I should add caching of CRL responses in this PR? It should be easy enough to do (key being the CRL endpoint URL).

adamdecaf:
I ran this locally and got it working. `revoked.badssl.com` looks like this:

```
--- Analyzers ---
* CAA records: not found
* Certificate: Revoked (at 2018-03-18 17:05:39 +0000 UTC)
```

adamdecaf:
@jvehent Regarding caching -- We wouldn't nessarily want a straight LRU cache. I assume we'd only want to keep CRL responses for some time (i.e. 1hour). That would refer more to a [Time aware least recently used (TLRU)](https://en.wikipedia.org/wiki/Cache_Replacement_Policies#Time_aware_least_recently_used_(TLRU)) cache. I didn't find one in the hashicorp library (and only [found one on github](https://github.com/thrisp/djinn/blob/master/cache.go). Thoughts?

We could spin up a goroutine to Purge the LRU cache every N minutes instead. Thoughts?

database/schema.sql
```diff
@@ -42,7 +42,8 @@ CREATE TABLE certificates(
     excluded_dns_domains        varchar[] NOT NULL DEFAULT '{}',
     excluded_ip_addresses       varchar[] NOT NULL DEFAULT '{}',
     is_technically_constrained  bool NOT NULL DEFAULT false,
-    cisco_umbrella_rank         integer NOT NULL DEFAULT 2147483647
+    cisco_umbrella_rank         integer NOT NULL DEFAULT 2147483647,
+    revoked_via_crl             bool NULL
```
jvehent: we already have `is_revoked bool` and `revoked_at timestamp` in the certificate table, could those be used instead of this new field?
I wonder if we need a `revocation_reason varchar` column to add extra information, like "listed in CRL XYZ and OCSP from ABC".

worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,200 @@
...
+	var wg sync.WaitGroup
+	var out [][]byte
+
+	crlPoints := cert.X509v3Extensions.CRLDistributionPoints
+	wg.Add(len(crlPoints))
+	for i := range crlPoints {
+		go func(point string, wg *sync.WaitGroup) {
+			defer wg.Done()
+
+			resp, err := httpClient.Get(point)
```
jvehent: how about using an http cache like https://github.com/gregjones/httpcache ?
note that I've never used it, so no idea if it'd work

worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,200 @@
...
+
+// AnalysisPrinter outputs results of comparing a certificate against the CRL(s) contained within
+func (eval) AnalysisPrinter(input []byte, printAll interface{}) (results []string, err error) {
+	var result Result
+	if err := json.Unmarshal(input, &result); err != nil {
+		return nil, fmt.Errorf("CRL Worker: failed to parse results: err=%v", err)
+	}
+
+	if result.Revoked {
+		results = append(results, fmt.Sprintf("* Certificate: Revoked (at %s)", result.RevocationTime.String()))
```
jvehent: nit: replace "Certificate" with "CRL", and remove the parenthesis around the timestamp

worker/crlWorker/crlWorker.go
```diff
@@ -0,0 +1,200 @@
...
+	} else {
+		results = append(results, "* Certificate: Not Revoked")
```
jvehent: nit: replace "Certificate" with "CRL"

jvehent:
I think caching for one hour is reasonable. Any method that achieves this is good. We run about 10k scans / hour right now due to the symantec top1m scanning, and we definitely can't download a fresh crl for each one of them.

jvehent:
I think you should also flip the `is_valid` bit in the `scan` table to false if the certificate is revoked. This is currently done by `tlsobs-scanner` when evaluating the chain of trust, but the worker could update the field since it runs later.

adamdecaf:
@jvehent this should be all fixed up now. There's no migration anymore, so we can skip that for now.

PR #312, opened 2018-03-18
connection: Print stderr from running cipherscan

adamdecaf:
I had an old version of bash, but didn't find out until running `cipherscan` manually.

### DIFF
--- a/connection/retriever.go
+++ b/connection/retriever.go
@@ -36,6 +36,7 @@ func Connect(domain, cipherscanbinPath string) ([]byte, error) {
 	comm.Stderr = &stderr
 	err := comm.Start()
 	if err != nil {
+		log.Println(stderr.String())
 		log.Println(err)
 		return nil, err
 	}


tomato42:
I wonder if it shouldn't be detected through return codes

adamdecaf:
Perhaps, but I was just getting `exit status 1`.

jvehent:
Perhaps only log if cipherscan returns a non-zero error code? That would reduce verbosity.

tomato42:
> Perhaps, but I was just getting `exit status 1`.

honestly I don't recall cipherscan code now, but if that `1` is not unique, maybe we should have a unique code that means "sanity of environment not established, either bash or openssl missing"

printing full cipherscan output on any error code would also work nicely

adamdecaf:
> printing full cipherscan output on any error code would also work nicely

I can print stdout too in this PR. 

I don't think `exit status 1` gives us anything, but we can keep printing errors on any failure.

PR #319, opened 2018-03-28
Update the names for ECDHE-ECDSA-CHACHA20-POLY1305 and ECDHE-RSA-CHACHA20-POLY1305 ciphers

cosmincojocar:
We use this mapping to dynamically generate the TLS rules in [gas](https://github.com/GoASTScanner/gas/) from the published `server-side-tls-conf.json` file.  It seems somehow that the name mapping for IANA isn't updated for these ciphers. Any reason for leaving the names empty?

We have this bug https://github.com/GoASTScanner/gas/issues/190 because of this.


jvehent:
I think we should keep both because they have different codes. The current ones are
* ECDHE-RSA-CHACHA20-POLY1305-OLD which hash code 0xCC13 or 52243
* ECDHE-ECDSA-CHACHA20-POLY1305-OLD which has code 0xCC14 or 52244
* DHE-RSA-CHACHA20-POLY1305-OLD which has code 0xCC15 or 52245

The new ones are
* ECDHE-RSA-CHACHA20-POLY1305 which has code 0xCCA8 or 52392
* ECDHE-ECDSA-CHACHA20-POLY1305 which has code 0xCCA9 or 52393
* DHE-RSA-CHACHA20-POLY1305 which has code 0xCCAA or 52394

cosmincojocar:
I am trying to map the cipher names `ECDHE-RSA-CHACHA20-POLY1305` and `ECDHE-ECDSA-CHACHA20-POLY1305` published in https://statics.tls.security.mozilla.org/server-side-tls-conf.json to the `IANAName` used in Go crypto/tls package. I used for this the map from `github.com/mozilla/tls-observatory/constants` package.  

Please could you point me to the duplicate? I don't see it in the `ciphersuites.go` file.

cosmincojocar:
@jvehent I added back the old definitions and updated the codes of the new definitions as suggested.

jvehent:
Awesome! Thanks for the patch.

Issue #316, opened 2018-03-25
OSX local dev improvements 


adamdecaf:
From the few larger changes I've submitted there are a few local changes I keep around. Are there thoughts on adding them? I can probably fit them all in so linux and osx dev work at the same time. 

This can likely be changed with a `if` inside the makefile. 
```
modified   Makefile
@@ -9,7 +9,7 @@ BUILDREV	:= $(BUILDDATE)+$(BUILDREF).$(BUILDENV)
 
 # Supported OSes: linux darwin windows
 # Supported ARCHes: 386 amd64
-OS			:= linux
+OS			:= darwin
```

I believe this is resolved by upgrading the vendored `crypto/x509` fork in this project, but I don't know if there are any diffs applied to our version. See https://github.com/zmap/zcrypto/issues/106 and https://github.com/google/certificate-transparency-go/commit/391726f8973ddb3521ab1a462bd55a35c6f67f0b 
```
modified   certificate/constraints/vendor/constraintcrypto/x509/root_cgo_darwin.go
@@ -157,7 +157,7 @@ import (
 func loadSystemRoots() (*CertPool, error) {
 	roots := NewCertPool()
 
-	var data C.CFDataRef = nil
+	var data C.CFDataRef
```

Lastly, I've mucked around with the `conf/*` files, but I don't think that's important to commit upstream. Maybe just exposing the `postgres` port onto the docker host.

jvehent:
I'm in favor of making development on macos easier, that said:

> This can likely be changed with a if inside the makefile.

You can override this on the command line by setting `OS=darwin make`

> I believe this is resolved by upgrading the vendored crypto/x509 fork in this project, but I don't know if there are any diffs applied to our version.

The one under `certificate/constraints/vendor/constraintcrypto/x509` has been customized and would need to be rebased.

> Lastly, I've mucked around with the conf/* files

You can easily point to a customized conf file with `-c`.

jvehent:
not at all. make a PR with updated deps and we'll merge it if CI is happy

PR #183, opened 2016-12-06
Added technical constraint checks

autrilla:
I've made some small changes to @jcjones's work on technical constraints and vendored them on certificate/constraints/vendor/crypto/x509, and I provided a small API over these specific changes that allows passing in a stdlib x509.Certificate and getting name constraints back.

A few notes:

- lib/pq does not support cidr types yet, so we have to store it as varchar[] and convert back and forth.
- I made some changes to certificate/constraints/vendor/crypto/x509's cgo symbol names, so they don't collide with the standard library's crypto/x509
- I have tested this with Deutsche Post's certificate, with www.intraship-dhl.at specifically.

### Warning: this contains changes to the DB (I've set up defaults for NOT NULL columns in schema.sql so nothing breaks).

PR #322, opened 2018-03-29
use upstream Go for technical constraint 

adamdecaf:
This change drops the `crypto/x509` fork for technically constrained certificates. Go1.10 (now required) adds those fields on `x509.Certificate`. 

The original constraints (and vendor fork) was added in https://github.com/mozilla/tls-observatory/pull/183. None of the constraint checking code needed to be changed. 

I see http://www.intraship-dhl.at/ as a test case, but I assume there are others to verify? (Example: https://github.com/mozilla/tls-observatory/pull/183#issuecomment-268531174)

This also resolves part of https://github.com/mozilla/tls-observatory/issues/316, which allows local dev on more platforms.


adamdecaf:
I created a diff from go's 1.9 release branch and the fork here. I'm going to go over it and try to find any customizations missing in go1.10. 

[x509.diff.txt](https://github.com/mozilla/tls-observatory/files/1861425/x509.diff.txt)

jvehent:
First off, this is really cool! Thanks for the patch.
I see there's a conflict in the vendoring. Maybe that folder should be removed entirely?

Also, upstream has more constraints parsed now. Should we add those to TLS Observatory as well?
```go
        // Name constraints
        PermittedDNSDomainsCritical bool // if true then the name constraints are marked critical.
        PermittedDNSDomains         []string
        ExcludedDNSDomains          []string
        PermittedIPRanges           []*net.IPNet
        ExcludedIPRanges            []*net.IPNet
        PermittedEmailAddresses     []string
        ExcludedEmailAddresses      []string
        PermittedURIDomains         []string
        ExcludedURIDomains          []string
```

jvehent:
Separate PR to add the new constraints makes sense to me.

adamdecaf:
I'm going to remove the vendor folder in a different change. Here's the first part. https://github.com/mozilla/tls-observatory/pull/333

PR #323, opened 2018-03-30
Remove marketplace* Bug 1432964

jasonthomas:
r?


jvehent:
cc @psiinon to do the same in our services metadata

PR #324, opened 2018-03-30
Add probeinfo.telemetry.mozilla.org Bug 1444942

jasonthomas:
r?


jvehent:
cc @psiinon to add to our services metadata

PR #326, opened 2018-03-30
Stop monitoring everything.me

bqbn:
The service is decommissioned per
* https://bugzilla.mozilla.org/show_bug.cgi?id=1432964
* https://bugzilla.mozilla.org/show_bug.cgi?id=1447049


bqbn:
@jasonthomas should `geodude.services.mozilla.com` be removed from the configs too? I see it's under `everything.me paging` section, but unlike the other two, the host is still resolvable.

Issue #338, opened 2018-04-21
Add ExtendedKeyUsagOID to the JSON return


christopher-henderson:
Currently when we construct the list of EKUs for a given certificate we translate known EKUs [to human readable strings ](https://github.com/mozilla/tls-observatory/blob/7acd6dd97ddbf2dc46fe6fe3b77b59c05bb60c1b/certificate/certificate.go#L293) (E.G. 1.3.6.1.5.5.7.3.9 --> "ExtKeyUsageOCSPSigning").

The folks at the CCADB would appreciate an _addition_ to the API which gives all raw OIDs for the EKUs alongside the translated slice of EKUs.

Issue #318, opened 2018-03-26
establish a database migration strategy


adamdecaf:
Right now keeping a postgres database for `tls-observatory` is an artisanal chore. We should automate it instead so as to not create issues for users deploying their own instances. 
 (See https://github.com/mozilla/tls-observatory/issues/317)

Personally I've used https://github.com/rubenv/sql-migrate before, but I'm sure there are a wealth of other options. [awesome-go offers a few other migration tools](https://github.com/avelino/awesome-go#database).

jvehent:
I'm ok with using a migration tool, especially if its automated (deploying the production service is fully automated so it'd be nice to not require manual commands).
If you already have experience with a specific tool, would you mind proposing a patch?

PR #339, opened 2018-04-21
Return raw OIDs for certificate EKUs

christopher-henderson:
This change extends the TLS Observatory to return a list of raw OIDs for EKUs alongside the translated list of EKUs

The end result looks something like...

```json
        "extendedKeyUsage": [
            "ExtKeyUsageClientAuth",
            "ExtKeyUsageEmailProtection",
            "ExtKeyUsageOCSPSigning",
            "1.3.6.1.5.5.7.3.14",
            "1.3.6.1.4.1.311.10.3.4",
            "1.3.6.1.4.1.311.10.3.4.1",
            "1.3.6.1.4.1.311.10.3.11",
            "1.3.6.1.4.1.311.20.2.1",
            "1.3.6.1.4.1.311.20.2.2",
            "1.3.6.1.4.1.311.21.19",
            "1.3.6.1.4.1.311.21.5",
            "1.3.6.1.4.1.311.21.6",
            "1.3.6.1.4.1.311.10.3.12"
        ],
        "extendedKeyUsageOID": [
            "1.3.6.1.5.5.7.3.2",
            "1.3.6.1.5.5.7.3.4",
            "1.3.6.1.5.5.7.3.9",
            "1.3.6.1.5.5.7.3.14",
            "1.3.6.1.4.1.311.10.3.4",
            "1.3.6.1.4.1.311.10.3.4.1",
            "1.3.6.1.4.1.311.10.3.11",
            "1.3.6.1.4.1.311.20.2.1",
            "1.3.6.1.4.1.311.20.2.2",
            "1.3.6.1.4.1.311.21.19",
            "1.3.6.1.4.1.311.21.5",
            "1.3.6.1.4.1.311.21.6",
            "1.3.6.1.4.1.311.10.3.12"
        ],
```

Addresses issue #338

### DIFF
--- a/certificate/certificate.go
+++ b/certificate/certificate.go
@@ -310,8 +310,7 @@ func GetValidityMap(trusted_ubuntu, trusted_mozilla, trusted_microsoft, trusted_
 
 }
 
-func getExtKeyUsages(cert *x509.Certificate) []string {
-	usage := make([]string, 0)
+func getExtKeyUsages(cert *x509.Certificate) (usage []string) {
 	for _, eku := range cert.ExtKeyUsage {
 		usage = append(usage, ExtKeyUsage[eku])
 	}
@@ -321,8 +320,7 @@ func getExtKeyUsages(cert *x509.Certificate) []string {
 	return usage
 }
 
-func getExtKeyUsageOIDs(cert *x509.Certificate) []string {
-	usage := make([]string, 0)
+func getExtKeyUsageOIDs(cert *x509.Certificate) (usage []string) {
 	for _, eku := range cert.ExtKeyUsage {
 		usage = append(usage, ExtKeyUsageOID[eku])
 	}
--- a/certificate/certificate.go
+++ b/certificate/certificate.go
@@ -94,6 +94,7 @@ type Extensions struct {
 	SubjectKeyId             string   `json:"subjectKeyId"`
 	KeyUsage                 []string `json:"keyUsage"`
 	ExtendedKeyUsage         []string `json:"extendedKeyUsage"`
+	ExtendedKeyUsageOID      []string `json:"extendedKeyUsageOID"`
 	SubjectAlternativeName   []string `json:"subjectAlternativeName"`
 	CRLDistributionPoints    []string `json:"crlDistributionPoint"`
 	PolicyIdentifiers        []string `json:"policyIdentifiers,omitempty"`
@@ -110,10 +111,10 @@ type X509v3BasicConstraints struct {
 }
 
 type Chain struct {
-	Domain string   `json:"domain"`
-	IP     string   `json:"ip"`
+	Domain string `json:"domain"`
+	IP     string `json:"ip"`
 	// base64 DER encoded certificates
-	Certs  []string `json:"certs"`
+	Certs []string `json:"certs"`
 }
 
 type IDs struct {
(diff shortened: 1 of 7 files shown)


certificate/certificate.go
```diff
@@ -110,10 +111,10 @@ type X509v3BasicConstraints struct {
 }
 
 type Chain struct {
-	Domain string   `json:"domain"`
```
christopher-henderson: This was `gofmt` autoformatting.

certificate/certificate.go
```diff
@@ -177,6 +178,25 @@ var ExtKeyUsage = [...]string{
 	"ExtKeyUsageOCSPSigning",
 	"ExtKeyUsageMicrosoftServerGatedCrypto",
 	"ExtKeyUsageNetscapeServerGatedCrypto",
+	"ExtKeyUsageMicrosoftCommercialCodeSigning",
```
christopher-henderson: Noticed that [x509 package](https://golang.org/src/crypto/x509/x509.go?s=20488:24535#L576) had these as further known EKUs.

database/schema.sql
```diff
@@ -140,3 +140,6 @@ GRANT SELECT ON analysis, certificates, scans, trust TO tlsobsscanner;
 GRANT INSERT ON analysis, certificates, scans, trust TO tlsobsscanner;
 GRANT UPDATE ON analysis, certificates, scans, trust TO tlsobsscanner;
 GRANT USAGE ON analysis_id_seq, certificates_id_seq, scans_id_seq, trust_id_seq TO tlsobsscanner;
+
+-- # Migrations
+ALTER TABLE certificates ADD COLUMN x509_extendedKeyUsageOID jsonb NULL;
```
jvehent: I think we should put this is a separate file under `database/migrations/` and label it with the release version (in this case, `1.3.3`)

certificate/certificate.go
```diff
@@ -301,6 +321,17 @@ func getExtKeyUsages(cert *x509.Certificate) []string {
 	return usage
 }
 
+func getExtKeyUsageOIDs(cert *x509.Certificate) []string {
+	usage := make([]string, 0)
```
jvehent: doesn't hurt, but also unnecessary. adding `(usage []byte)` as a return in the function signature could get rid of this.

database/schema.sql
adamdecaf: Are you thinking of putting manual migrations there until #318 is fixed?

certificate/certificate.go
autrilla: If you're initializing this with `make`, set the initial size to `len(cert.ExtKeyUsage) + len(cert.UnknownExtKeyUsage)`, or do what @jvehent suggests.

database/certificate.go
```diff
@@ -29,6 +29,11 @@ func (db *DB) InsertCertificate(cert *certificate.Certificate) (int64, error) {
 		return -1, err
 	}
 
+	extkeyusageOID, err := json.Marshal(cert.X509v3Extensions.ExtendedKeyUsageOID)
```
autrilla: nit: consistent naming, if you're using `ExtKeyUsageOID` for the struct, you should use `extKeyUsageOID` here.

database/certificate.go
```diff
@@ -184,6 +191,11 @@ func (db *DB) UpdateCertificate(cert *certificate.Certificate) error {
 		return err
 	}
 
+	extkeyusageOID, err := json.Marshal(cert.X509v3Extensions.ExtendedKeyUsageOID)
```
autrilla: Same as above

database/certificate.go
```diff
@@ -452,11 +467,11 @@ type Scannable interface {
 func (db *DB) scanCert(row Scannable) (certificate.Certificate, error) {
 	cert := certificate.Certificate{}
 
-	var crl_dist_points, extkeyusage, keyusage, subaltname, policies, issuer, subject, key []byte
+	var crl_dist_points, extkeyusage, extkeyusageOID, keyusage, subaltname, policies, issuer, subject, key []byte
```
autrilla: Same nit as above

christopher-henderson: Yeah, I'm aware of Go's convention. This was so named to stay inline with the surrounding names that already existed ("when in Rome..."). I can still change these if you'd like, or do a general refactor of the names in the package for a little bit of spring cleaning.

christopher-henderson:
Right, the tests don't know where to find the migration files now. I'll have to update that.

jvehent:
Sorry, wasn't clear: you should update the schema to create the column AND
add a migration file for people with existing DBS.

jvehent:
This looks good, thanks @christopher-henderson ! I'll get it deployed with the next release.

PR #344, opened 2018-05-08
Fix creation of EKU OID column

jvehent:
When adding new columns, existing certs must be updated to prevent the parsing to fail.


christopher-henderson:
Good catch.

As an aside, when a long term migration scheme gets figured out it may be prudent to bundle a rescan script into the procedure.

PR #348, opened 2018-05-12
Added tool to initialize the mozilla policy field

christopher-henderson:
There was a wild fire which came up to the property and forced a short evacuation, so I didn't get to test this.

### DIFF
--- a/tools/initMozPolicy.go
+++ b/tools/initMozPolicy.go
@@ -88,7 +88,7 @@ func main() {
 			if !first {
 				sql += ","
 			}
-			sql += fmt.Sprintf("(%d, '%s')", id, mozPolicy)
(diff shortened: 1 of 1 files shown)


tools/initMozPolicy.go
```diff
@@ -0,0 +1,102 @@
...
+			break
+		}
+		// batch update
+		sql := "UPDATE certificates SET mozillaPolicyV2_5 = newvalues.mozPolicy FROM ( VALUES "
+		first := true
+		for id, mozPolicy := range updates {
+			if !first {
+				sql += ","
+			}
+			sql += fmt.Sprintf("(%d, '%s')", id, mozPolicy)
```
jvehent: The `jsonb` type assertion is required here, correct code is:
```go
sql += fmt.Sprintf("(%d, '%s'::jsonb)", id, mozPolicy)
```

Issue #144, opened 2016-09-16
add a field to aid in identifying CAs by name/spki


mozkeeler:
Sometimes CAs have multiple different certificates that nonetheless represent the same logical CA. It would be nice to have a field like 'certID' that is something like SHA-256(der of name + der of spki).

jvehent:
We have a field with the public key hash. Would that fulfill your use case?

mozkeeler:
Well, we could probably implement it in terms of that, but I think the idea is that it would be a single identifier that would make these two CAs identical:

{ issuer: "Foocorp Root CA", subject: "Foocorp Root CA", spki: [some key], ...}
{ issuer: "Cross-sign CA", subject: "Foocorp Root CA", spki: [same key], ...}

but these would be distinct:

{ issuer: "Foocorp Root CA", subject: "Foocorp Root CA", spki: [some key], ...}
{ issuer: "Foocorp Root CA", subject: "Foocorp Root CA-Bv3 Final Final", spki: [same key], ...}

(since for a certificate to be issued by another one, the issuer/subject have to match)

jvehent:
That seems reasonable. Could you share some pseudo code (or, even better, a patch) ?
That would most likely go in the cert parsing code here: https://github.com/mozilla/tls-observatory/blob/master/certificate/certificate.go#L406-L471

sleevi:
@jvehent 

```go
func CertIDHash(cert* x509.Certificate) string {
  h := sha256.New();
  h.Write(cert.RawSubject);
  h.Write(cert.RawSubjectPublicKeyInfo);
  return base64.StdEncoding.EncodeToString(h.Sum(nil));
```

jvehent:
This is live:
```bash
$ curl -s https://tls-observatory.services.mozilla.com/api/v1/certificate?id=42 | jq -r '.hashes.sha256_subject_spki'
065D43F4C81B18BE6989435AEA98144938DB896DC9EB6573BF219CDEC2AF6D80
```
The backfill of the certificates table will take a couple days, as the script needs to update the sha256_subject_spki field in all 2M certs in the DB.

PR #353, opened 2018-06-01
Re-add support for Subject SPKI hash

jvehent:
In 1.2.37, I removed support for Subject SPKI (#144) and turned the value into standard SPKI. Turns out this was a mistake, and Subject SPKI is very much needed by CCADB. This patch re-adds it into a new column.

## Checklist
* [x] Run `make`, `gofmt` and `golint` your code, and run a test scan on your local machine before submitting for review.
* [x] Workers needs an AnalysisPrinter, registered via `worker.RegisterPrinter()` (which is separate from `worker.RegisterWorker()`), and imported in [tlsobs](https://github.com/mozilla/tls-observatory/blob/master/tlsobs/main.go#L20-L28) ([example](https://github.com/mozilla/tls-observatory/blob/master/worker/awsCertlint/awsCertlint.go#L39-L56)).
* [x] When adding new columns to the database, also add a DB migration script under `database/migrations` named the **next** release (eg. if current release is 1.3.2, migration file will be `1.3.3.sql`).
* [x] When new columns require data to be recomputed, add a script under `/tools` ([example](https://github.com/mozilla/tls-observatory/blob/master/tools/fixSHA256SubjectSPKI.go)) that updates the database and will be run by administrators.

### DIFF
--- a/database/schema.sql
+++ b/database/schema.sql
@@ -2,6 +2,7 @@ CREATE TABLE certificates(
     id                          serial primary key,
     sha1_fingerprint            varchar NOT NULL,
     sha256_fingerprint          varchar NOT NULL,
+    sha256_spki                 varchar NOT NULL,
     sha256_subject_spki         varchar NOT NULL,
     pkp_sha256                  varchar NOT NULL,
     serial_number               varchar NULL,
--- a/certificate/certificate.go
+++ b/certificate/certificate.go
@@ -61,11 +61,12 @@ type MozillaPolicy struct {
 }
 
 type Hashes struct {
-	MD5        string `json:"md5,omitempty"`
-	SHA1       string `json:"sha1,omitempty"`
-	SHA256     string `json:"sha256,omitempty"`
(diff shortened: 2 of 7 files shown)


database/schema.sql
```diff
@@ -141,4 +142,4 @@ ALTER ROLE tlsobsscanner WITH NOSUPERUSER INHERIT NOCREATEROLE NOCREATEDB LOGIN
 GRANT SELECT ON analysis, certificates, scans, trust TO tlsobsscanner;
 GRANT INSERT ON analysis, certificates, scans, trust TO tlsobsscanner;
 GRANT UPDATE ON analysis, certificates, scans, trust TO tlsobsscanner;
-GRANT USAGE ON analysis_id_seq, certificates_id_seq, scans_id_seq, trust_id_seq TO tlsobsscanner;
\ No newline at end of file
+GRANT USAGE ON analysis_id_seq, certificates_id_seq, scans_id_seq, trust_id_seq TO tlsobsscanner;
```
jcjones: New /extra newline here OK?

PR #376, opened 2019-01-17
Updates for services bpitts works on

sciurus:
I realized I'd never updates this as domains were added.


jvehent:
Do you want paging as well? That would be done by duplicating these hosts in a block further down, with the paging email.

sciurus:
Thanks for mentioning that! I'll update this PR.

sciurus:
I've added the paging config.

PR #379, opened 2019-02-11
add lando

ckolos:
who's got two thumbs and almost let this cert expire? >This guy<


conf/runner.yaml
```diff
@@ -49,6 +49,10 @@ runs:
         - fxpartners.mozilla.com
 
         # Conduit
+        - lando.devsvcdev.mozaws.net
+        - lando.services.mozilla.com
+        - api.lando.devsvcdev.mozaws.net
+        - api.lando.services.mozilla.com
```
jvehent: do you want a paging conf for those? If so, that needs to be added further down the yaml file

PR #385, opened 2019-03-13
Remove download-sha1.cdn.mozilla.net

jbuck:
download-sha1.cdn.mozilla.net is expiring in 2 weeks and not being renewed


jbuck:
@oremj @jvehent

PR #401, opened 2020-02-07
update truststores

g-k:
ran `make truststores`


g-k:
This was just a test to see if it updating the truststores would fix CI. I think we're still failing to find update CA stores for Microsoft.

PR #404, opened 2020-03-03
Socorro: remove crash-reports-xpsp2

sciurus:
We are retiring this endpoint; see https://bugzilla.mozilla.org/show_bug.cgi?id=1614692


sciurus:
What's the right process to ask for this to be deployed?

Until that's done we're going to get spammed with errors in the TLS Observatory runner results emails.

PR #424, opened 2021-06-09
Remove monitring for elmo

sciurus:
l10n.mozilla.org was decommed a while back. This is now a redirect that
lives in refractr. It is monitored by Web SRE and should not page
anyone.


g-k (approved):
r+ pending green CI

PR #442, opened 2024-07-01
Disable Shavar certificate notifications.

smarnach:
If the Shavar production certificate has less than two weeks to go, TLS Observatory currently still pages a legacy service in the wrong PagerDuty account. This PR simply disables the notification so I no longer get paged for it.


smarnach:
CC @mozilla/websre

